PR #13, opened 2016-01-07
[MNG-5958] restore binary compatibility of Lifecycle.setPhases

atanasenko:


michael-o:
Thank you. I will have a look and run it.

atanasenko:
btw, I'm also seeing a failure in test for MNG-4919, which seems to be caused by MNG-5359

michael-o:
Just trying to run the ITs without your patch first.

michael-o:
Seems like I can't run the tests in a locked down environment, regardless of a Nexus instance and am HTTP proxy. Artifact resolution always points to `file:target/null`

atanasenko:
I've had troubles running tests when following official guidelines [1].
But changing 'mvn clean test ...' to 'mvn clean install ...' helped.
[1] https://maven.apache.org/core-its/core-it-suite/

michael-o:
The Its do not work if you are behind a repo manager and in an locked down environment, regardless of the tips on the site.

jvanzyl:
Once all the files are in the bootstrap that you might need I've never had an issue. I'll double check on Monday, but last week I ran the ITs on a machine with only access to our Nexus proxy. I've basically had the same local setup (across various versions of OSX) for 8 years so I may well have something here that you don't have. But I haven't had any accessibility problems for a long time.

michael-o:
That is strange, I ran the ITs from my Windows and FreeBSD machine at work. Both failed with `maven.it.central` pointing to our Nexus instance. On my home machine it didn't work either unless I have removed the mirror from my local `settings.xml`.

PR #21, opened 2017-04-12
[MNG-6127] Add plugin execution configuration interference test

mkrizmanic:


core-it-suite/src/test/resources/mng-6127/plugin/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>mng-6127</groupId>
+  <artifactId>mng-6127-plugin</artifactId>
+  <version>0.1</version>
+  <packaging>maven-plugin</packaging>
+
+  <properties>
+    <maven-version>3.3.1</maven-version>
```
michael-o: This should rather be `maven.version` as we always do in POMs.

mkrizmanic: Hi, it should be defined here because we do not have a parent pom which contains the next artifact versions.

The same pattern is used in other ITs. For instance:

https://github.com/apache/maven-integration-testing/blob/master/core-it-suite/src/test/resources/mng-5578-session-scope/plugin/pom.xml

michael-o: Alright, my bad. Thanks for the clarification.

mkrizmanic:
already merged

PR #22, opened 2017-07-05
[MNG-5965] Parallel build multiplies work if multiple goals are given

dbmeneses:


michael-o:
@dbmeneses why is there no parent element on the module POMs?

PR #23, opened 2017-07-17
[MNG-6265] Integration test for Maven script can break if "-f" path contains special characters 

etzelc:
Add the integration test for issue MNG-6256 (pull request apache/maven#128). https://issues.apache.org/jira/browse/MNG-6256

I set the version contraint to `"[3.5.1,)"`.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
@@ -106,6 +106,7 @@ public class IntegrationTestSuite
         // -------------------------------------------------------------------------------------------------------------
         // suite.addTestSuite( MavenIT0108SnapshotUpdateTest.class ); -- MNG-3137
 
+        suite.addTestSuite( MavenITmng6256SpecialCharsAlternatePOMLocation.class );
         suite.addTestSuite( MavenITmng6240PluginExtensionAetherProvider.class );
         suite.addTestSuite( MavenITmng6223FindBasedir.class );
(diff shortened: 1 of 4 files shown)


Tunaki:
I ran the IT with 3.5.0 on Windows, but it passed. Since the issue resolves around the error message that is printed on the console, I think the two files `folder-with- -space/pom.xml` and `folder-with-)-closing-bracket/pom.xml` should be removed, and instead it should be tested that the correct error message is printed, right?

etzelc:
Thanks for your testing effort. I am surprised both tests pass in your Windows environment with 3.5.0. I tested the IT with two different clients (Windows 7 and Windows 10) and while `folder-with- -space/pom.xml` passes, the `folder-with-)-closing-bracket/pom.xml` test fails. 
The `folder-with- -space/pom.xml` test was developed to cover MNG-6177 (https://issues.apache.org/jira/browse/MNG-6177), which has no IT and was already targeting "special" characters like spaces and more.

I have implemented the IT in a way to perform a successful maven run, since this is what you usually expect from maven. The example `mvn -f "folderWithClosing)Bracket/pomFolder"` from issue MNG-6256 (https://issues.apache.org/jira/browse/MNG-6256) was chosen to have a very short example with no prerequisites. With 3.5.0 you should see the same error even if a well formed POM exist at the specified location and you set additionally a goal (see IT test `folder-with-)-closing-bracket/pom.xml`).
I did not implement a dedicated test for the error message, because it may change over the time and the purpose of my patch was not to have a correct error message (which is included in the patch implicitly), but have a successful maven run.

PR #24, opened 2017-07-18
Add test for MNG-6255 jvm.config line endings

grkvlt:
Integration test for fix made in https://github.com/apache/maven/pull/127


grkvlt:
Tested on OSX; passes when change from https://github.com/apache/maven/pull/127 are made to the `mvn` script.

grkvlt:
Removed *CR* test

PR #41, opened 2019-01-05
[MNG-5693] Remove check for link to cwiki

slachiewicz:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5482AetherNotFoundTest.java
```diff
@@ -90,7 +90,16 @@ public void check( String dir )
         assertTrue( "ClassNotFoundException message was not found in output.", msg >= 0 );
 
         int url = indexOf( lines, ".*http://cwiki.apache.org/confluence/display/MAVEN/AetherClassNotFound.*" );
-        assertTrue( "Url to ClassNotFoundAether was not found in output.", url >= 0 );
+
+        boolean versionWithMng5693 = matchesVersionRange( "[3.6.1,)"); // no more link to cwiki
```
Tibor17: Why we keep this waiting for 8 months?
The hint was printed by older Maven?

michael-o: Why not change this in `super()`? The test will be obsolete with 3.6.2?!

Tibor17: I think he is checking that next versions do not have that link either. If it is @slachiewicz 's purpose, then `super()` would not help.
@michael-o Question I asked on ML, why do we keep ITs for old versions. We don't need them. We are tesing SNAPSHOT version and we do it backwards with versions. Old versions are over, and there I would propose to TAG this project with every Maven release. Then delete non-general ITs.

PR #43, opened 2019-03-24
Add regression test for MNG-6506

veithen:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6506PackageAnnotationTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+        verifier = newVerifier( pluginDir.getAbsolutePath(), "remote" );
+        verifier.executeGoal( "install" );
+        verifier.resetStreams();
+        verifier.verifyErrorFreeLog();
+
+        verifier = newVerifier( projectDir.getAbsolutePath(), "remote" );
+        verifier.executeGoal( "verify" );
+        verifier.resetStreams();
+        verifier.verifyErrorFreeLog();
```
eolivelli: Maybe we could add a more narrow assertion, like printing something in the test case and then checking that the output contains such text

veithen: Done.

PR #44, opened 2019-07-03
MNG-5577 avoid false positives for different errors

josephw:
I ran into a false positive with this test while investigating MNG-5577. The build was failing with an `NPE`, which was enough to make it pass. It should fail due to an unavailable exception. This change checks that the build fails for the expected reason.

Also, a typo fix.


michael-o:
Running ITs

PR #46, opened 2019-08-14
[MNG-6731] Jetty getLocalPort() returns -1 resulting in build failures

Tibor17:
Jenkins build
https://builds.apache.org/job/maven-box/job/maven/job/MNG-6731/


michael-o:
Didn't you merge this already?

Tibor17:
not yet. I was watching TV ;-)
Let me merge it. Thx

Tibor17:
No. At least it was not me!

core-it-suite/src/test/java/org/apache/maven/it/HttpServer.java
```diff
@@ -66,21 +68,20 @@ public void stop()
...
+        server.setConnectors( new Connector[]{ new ServerConnector( server ) } );
         server.addBean( new ScheduledExecutorScheduler() );
 
-        ServerConnector http = new ServerConnector( server );
-        http.setPort( port );
-        http.setIdleTimeout( 30000 );
-        server.addConnector( http );
+        ServerConnector connector = (ServerConnector) server.getConnectors()[0];
+        connector.setIdleTimeout( 30_000L );
+        connector.setPort( port );
```
joakime: Why not just do this instead ?

``` java
ServerConnector connector = new ServerConnector( server );
connector.setIdleTimeout( 30_000L );
connector.setPort( port );
server.addConnector( connector );
```

core-it-suite/src/test/java/org/apache/maven/it/MavenIT0146InstallerSnapshotNaming.java
```diff
@@ -59,15 +60,11 @@ protected void setUp()
...
         {
-            if ( server.isFailed() )
-            {
-                fail( "Couldn't bind the server socket to a free port!" );
-            }
-            Thread.sleep( 100L );
+            fail( "Couldn't bind the server socket to a free port!" );
         }
-        port = server.getConnectors()[0].getLocalPort();
+        port = ( (NetworkConnector) server.getConnectors()[0] ).getLocalPort();
```
joakime: There's also `server.getURI().getPort()` as a valid option here.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0553SettingsAuthzEncryptionTest.java
```diff
@@ -86,18 +93,13 @@ protected void setUp()
         handlerList.addHandler( repoHandler );
         handlerList.addHandler( new DefaultHandler() );
 
-        server = new Server( 0 );
         server.setHandler( handlerList );
         server.start();
-        while ( !server.isRunning() || !server.isStarted() )
+        if ( server.isFailed() )
```
joakime: Server could have successfully started up, and then be successfully stopped, before this line hits.
In this scenario, the server LifeCycle state is STOPPED, not FAILED.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0768OfflineModeTest.java
```diff
@@ -103,17 +103,12 @@ else if ( request.getRequestURI().endsWith( ".md5" ) || request.getRequestURI().
         try
         {
             server.start();
-            while ( !server.isRunning() || !server.isStarted() )
+            if ( server.isFailed() )
```
joakime: This state is not possible to reach immediately after `server.start()` in a Bind failure scenario.

The ServerConnector's doStart will attempt to bind
The BindException gets thrown for failure to bind
A IOException wraps failure to bind during ServerConnector doStart
The Server doStart captures the IOException and adds it to the MultiException its tracking
At the end of Server doStart the MultiException throws if it has any exceptions its tracking.

This scenario should result in a IOException or MultiException from server.start()

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2305MultipleProxiesTest.java
```diff
@@ -71,20 +78,16 @@ public void testit()
...
-                fail( "Couldn't bind the server socket to a free port!" );
-            }
-            Thread.sleep( 100L );
+            fail( "Couldn't bind the server socket to a free port!" );
         }
-        int httpPort = server.getConnectors()[0].getLocalPort();
+        int httpPort = ( (NetworkConnector) server.getConnectors()[0] ).getLocalPort();
         System.out.println( "Bound server socket to HTTP port " + httpPort );
-        int httpsPort = server.getConnectors()[1].getLocalPort();
+        int httpsPort = ( (NetworkConnector) server.getConnectors()[1] ).getLocalPort();
```
joakime: The `HttpConfiguration` belonging to Connector[0] needs to have its `HttpConfiguration.setSecurePort(int)` set for the port that this HTTPS connector.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0553SettingsAuthzEncryptionTest.java
Tibor17: because this has a purpose. I want to catch all errors. Not only IOException.
```
    @Override
    public final void start() throws Exception
    {
        synchronized (_lock)
        {
            try
            {
                if (_state == __STARTED || _state == __STARTING)
                    return;
                setStarting();
                doStart();
                setStarted();
            }
            catch (Throwable e)
            {
                setFailed(e);
                throw e;
            }
        }
    }
```

joakime: That is the generic `start()` for `AbstractLifeCycle`.
`Server` is a `ContainerLifeCycle` and a special one at that.
It's `Server.doStart()` has to do far more then this abstract base behavior simply because it's a combination of a Container (with child lifecycles) and it's the main entry point for external that care about the LifeCycle state.

core-it-suite/src/test/java/org/apache/maven/it/MavenIT0146InstallerSnapshotNaming.java
Tibor17: I did not want to make the code too much different from what we had before.I was glad that the version 9.0.4 works. After I changed the version to 9.2.28 I got HTTP 405 and really did not want to get new arguments from the devs saying that there is something which broke new version.

Tibor17: We can improve as always but I spent 3 days with these 38 tests and for me it is always good to have several smaller changes than one big change. With small changes/commits we can always find test older commits if we find some issue.

Tibor17: So I would make a new commit as an improvement of this line.

Tibor17: @joakime 
Do you see more occurences of this? Notice that we also have two connectors in few tests using HTTPS.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0553SettingsAuthzEncryptionTest.java
Tibor17: I think the purpose of this is to check server's status in a parallel thread. ok, we can avoid the IF statement.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2305MultipleProxiesTest.java
Tibor17: The thing is that we do not set the port, via `HttpConfiguration.setSecurePort(int)`, because we do not know it. We can only read both ports as local ports.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0553SettingsAuthzEncryptionTest.java
joakime: If you have parallel processes sure.
But your usages are sequential.
If you decided to use something like DeploymentManager to scan for webapps and deploy them, then you would be interested in ThrowUnavailableOnStartup, which waits for initial deployment to report failures to Server.start

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2305MultipleProxiesTest.java
joakime: Then your Requests never indicate that they are secure.
The Requests will not contain the security details.
The Requests will not have the javax attributes for the SSL / Certificate state of the connection.
Nor will you have ability to use the "confidential" constraints.
And usages that redirect from HTTP to HTTPS will also fail. (such as SecureRedirectHandler)

Tibor17:
@joakime 
The best idea would to have a new PR with commits for each problem separately. Would you do that?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2305MultipleProxiesTest.java
Tibor17: @joakime
I am convinced your advice is right. Can you improve these tests and publish a new PR?

PR #47, opened 2019-08-31
[MNG-6357] Dependency order should be nearest first

belingueres:
- Added IT showing the functionality in MNG-6357;.
- Added upper bound to 3.6.2 in
MavenITmng3813PluginClassPathOrderingTest since its behavior changed
because of MNG-6357.


jmhodges:
Would love to see this updated, too! The conflict here (like https://github.com/apache/maven/pull/281 ) is real small and obvious. Happy to make a new PR if that's easier!

michael-o:
I think this only is valid for Maven 4.

PR #49, opened 2019-10-03
[MNG-6772] Prove Super POM central entry won't override

ewiegs4:
Integration tests for https://github.com/apache/maven/pull/289

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6772NestedImportScopeRepositoryOverride.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6772NestedImportScopeRepositoryOverride.java
@@ -1,0 +1,95 @@
+package org.apache.maven.it;
+
+import java.io.File;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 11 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6772NestedImportScopeRepositoryOverride.java
```diff
@@ -0,0 +1,95 @@
...
+ * The test confirms that the dominant repository definition (child) wins while resolving the import POMs.
+ *
+ */
+public class MavenITmng6772NestedImportScopeRepositoryOverride
+    extends AbstractMavenIntegrationTestCase
+{
+
+    public MavenITmng6772NestedImportScopeRepositoryOverride()
+    {
+        super( "(3.6.2,)" );
```
rfscholte: If this is a regression, please change it to `super( "(,3.6.2),(3.6.2,)" );` so we can verify it with older versions of Maven.

ewiegs4: It is not a regression that I am aware of. The tests fail on all released 3.5.x and 3.6.x versions. I can't guarantee that there isn't a 3.3.x version or older where this may work, but seems like a lot of effort to figure out the answer.

PR #50, opened 2019-10-12
[MNG-6759] - Add test demonstrating the issue where the wrong reposit…

srdo:
…ories are used for resolving transitive dependencies

Test for https://github.com/apache/maven/pull/294.

Advice on improving these would be appreciated. It's not good to be depending on an external plugin or an external repository.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6759TransitiveDependencyRepositoriesTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6759TransitiveDependencyRepositoriesTest.java
@@ -1,0 +1,64 @@
+package org.apache.maven.it;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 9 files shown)


srdo:
Fixed so the test now has no external dependencies.

rfscholte:
Don't forget to add your test to `org.apache.maven.it.IntegrationTestSuite`

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6759TransitiveDependencyRepositoriesTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * This is a test for <a href="https://issues.apache.org/jira/browse/MNG-6759">MNG-6759</a>.
+ */
+public class MavenITmng6759TransitiveDependencyRepositoriesTest extends AbstractMavenIntegrationTestCase {
+
+    private final String projectBaseDir = "/mng-6759-transitive-dependency-repositories";
+
+    public MavenITmng6759TransitiveDependencyRepositoriesTest() {
+        super( "(,3.6.1],[3.6.3,)" );
```
rfscholte: `super( "(,3.6.2),(3.6.2,)" );` is the preferred call.

srdo:
Thanks for letting me know about the suite. Addressed your comments.

PR #53, opened 2020-01-29
[MNG-4660] Add integration test to show that --resume-from works

mthmulders:
Will offer this one for merging once the bug has indeed been fixed, and the `// TODO` regarding the version range has been resolved.


mthmulders:
Superseeded by #54.

PR #54, opened 2020-01-29
[MNG-4660] Add integration test to show that --resume-from works

mthmulders:
Will offer this one for merging once the bug has indeed been fixed, and the `// TODO` regarding the version range has been resolved.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
@@ -108,6 +108,7 @@ public class IntegrationTestSuite
         // suite.addTestSuite( MavenIT0108SnapshotUpdateTest.class ); -- MNG-3137
 
         suite.addTestSuite( MavenITmng4660ResumeFromTest.class );
+        suite.addTestSuite( MavenITmng4660OutdatedPackagedArtifact.class );
         suite.addTestSuite( MavenITmng6759TransitiveDependencyRepositoriesTest.class );
         suite.addTestSuite( MavenITmng6720FailFastTest.class );
         suite.addTestSuite( MavenITmng6558ToolchainsBuildingEventTest.class );
--- a/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
@@ -107,6 +107,7 @@ public class IntegrationTestSuite
         // -------------------------------------------------------------------------------------------------------------
         // suite.addTestSuite( MavenIT0108SnapshotUpdateTest.class ); -- MNG-3137
 
+        suite.addTestSuite( MavenITmng4660ResumeFromTest.class );
         suite.addTestSuite( MavenITmng6759TransitiveDependencyRepositoriesTest.class );
         suite.addTestSuite( MavenITmng6720FailFastTest.class );
         suite.addTestSuite( MavenITmng6558ToolchainsBuildingEventTest.class );
@@ -234,7 +235,6 @@ public class IntegrationTestSuite
         suite.addTestSuite( MavenITmng4679SnapshotUpdateInPluginTest.class );
         suite.addTestSuite( MavenITmng4677DisabledPluginConfigInheritanceTest.class );
         suite.addTestSuite( MavenITmng4666CoreRealmImportTest.class );
-        suite.addTestSuite( MavenITmng4660ResumeFromTest.class );
         suite.addTestSuite( MavenITmng4654ArtifactHandlerForMainArtifactTest.class );
         suite.addTestSuite( MavenITmng4644StrictPomParsingRejectsMisplacedTextTest.class );
         // suite.addTestSuite( MavenITmng4633DualCompilerExecutionsWeaveModeTest.class );
--- a/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
@@ -234,6 +234,7 @@ public class IntegrationTestSuite
         suite.addTestSuite( MavenITmng4679SnapshotUpdateInPluginTest.class );
         suite.addTestSuite( MavenITmng4677DisabledPluginConfigInheritanceTest.class );
         suite.addTestSuite( MavenITmng4666CoreRealmImportTest.class );
+        suite.addTestSuite( MavenITmng4660ResumeFromTest.class );
         suite.addTestSuite( MavenITmng4654ArtifactHandlerForMainArtifactTest.class );
         suite.addTestSuite( MavenITmng4644StrictPomParsingRejectsMisplacedTextTest.class );
         // suite.addTestSuite( MavenITmng4633DualCompilerExecutionsWeaveModeTest.class );
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
@@ -27,6 +27,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
+
 /**
  * This is a test case for a new check introduced with <a href="https://issues.apache.org/jira/browse/MNG-4660">MNG-4660</a>.
  * That check verifies if a packaged artifact within the Reactor is up-to-date with the outputDirectory of the same project.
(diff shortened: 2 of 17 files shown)


core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
```diff
@@ -231,6 +231,7 @@ public static Test suite()
         suite.addTestSuite( MavenITmng4679SnapshotUpdateInPluginTest.class );
         suite.addTestSuite( MavenITmng4677DisabledPluginConfigInheritanceTest.class );
         suite.addTestSuite( MavenITmng4666CoreRealmImportTest.class );
+        suite.addTestSuite( MavenITmng4660ResumeFromTest.class );
```
rfscholte: In general this is in order of implementation, not in order of reporting (also helps with fast feedback)

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
```diff
@@ -0,0 +1,74 @@
...
+        finally
+        {
+            verifier1.resetStreams();
+        }
+
+        final Verifier verifier2 = newVerifier( testDir.getAbsolutePath() );
+        verifier2.setAutoclean( false );
+        verifier2.addCliOption( "--resume-from" );
+        verifier2.addCliOption( ":module-b" );
+        verifier2.executeGoal( "compile" ); // to prevent the unit test from failing (again)
```
rfscholte: This looks like a mixup of 2 separate issues:
1. you want to continue and reuse the artifacts that have already been build. In that case the goal should be `verify`.
2. you want to compile module-b, but since Maven hasn't attached its artifacts to the reactor, you need to recompile module-a too. In that case I would add `verifier2.addCliOption( "--also-make" );`

Both are interesting and don't work right now. 1 is harder (and matches the description of the issue), 2 should be reasonably easier and already a huge improvement, but requires a separate issue.

rfscholte: BTW I do like this usecase: ignore the testresult of an early module for now, continue with its artifact to see the result for the rest of the project.

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
mthmulders: [Line 723](https://github.com/apache/maven-integration-testing/blob/e8f24ba23d9a8d97a2e1e8c9920213ffb6ebe772/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java#L723) suggests something else...:

>  Add tests in reverse alpha order above.

rfscholte: I see, and https://github.com/apache/maven-integration-testing/blob/e8f24ba23d9a8d97a2e1e8c9920213ffb6ebe772/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java#L88-L93 seems contradictionary. This text must be from the period we didn't start with MNG-nnn yet. However, the intention is right: newer tests are more likely to fail, so start with those.
I'll rewrite it...

mthmulders: OK, I'll await the outcome of that and then update the place of this new test in the suite.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
mthmulders: We do option 2, but instead of recompiling module-a, we want to reuse the outcome of the previous build, which did include module-a. That's why we have `--resume-from :module-b` and _autoclean_ set to off. The test that resides in module-b would fail again and that's why we went for `compile`.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * This is a test case for <a href="https://issues.apache.org/jira/browse/MNG-4660">MNG-4660</a>.
+ *
+ * @author Maarten Mulders
+ * @author Martin Kanters
+ */
+public class MavenITmng4660ResumeFromTest extends AbstractMavenIntegrationTestCase {
+    public MavenITmng4660ResumeFromTest()
+    {
+        super( "[3.7.0-SNAPSHOT,)" );
```
rfscholte: This should be 3.7.0, the  `AbstractMavenIntegrationTestCase` understands that you need to test this a SNAPSHOT.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6090CIFriendlyTest.java
```diff
@@ -61,7 +60,7 @@ public void testitShouldResolveTheDependencies()
         
         verifier.addCliOption( "-Drevision=1.2" );
         verifier.setLogFileName( "install-log.txt" );
-        verifier.executeGoals( Arrays.asList( "clean", "install" ) );
```
rfscholte: Please don't change other integration tests, only the version-range if the behavior changes (or add code)
I'm pretty sure you want to test the `install` phase when working with CI-friendly versions, to ensure that still works.

MartinKanters: Right, I'm not sure if this was an intended change. I've reverted it.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
MartinKanters: Interesting! Fixed.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6090CIFriendlyTest.java
mthmulders: It was an intended change, we figured that `clean install` was no longer needed. But I guess it'll still work 😄

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
MartinKanters: I've put the test at the top of the list. Do you want us to change the javadoc (pointed out by Maarten above) or do you want to do that separately?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
```diff
@@ -0,0 +1,91 @@
...
+
+        verifier1.verifyErrorFreeLog();
+        verifier1.resetStreams();
+
+
+        // 2. Create a properties file with some content and compile only that module (module A).
+        final Verifier verifier2 = newVerifier( testDir.getAbsolutePath() );
+
+        final Path resourcesDirectory = Files.createDirectories( Paths.get( testDir.toString(), "module-a", "src", "main", "resources" ) );
+        System.err.println( "Created directory " + resourcesDirectory.toAbsolutePath().toString() );
```
rfscholte: System.x.println don't belong in tests

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
```diff
@@ -0,0 +1,91 @@
...
+        final Path fileToWrite = resourcesDirectory.resolve( "example.properties" );
+        FileUtils.fileWrite( fileToWrite.toString(), "x=42" );
+        System.err.println( "Wrote file " + fileToWrite.toAbsolutePath().toString() );
```
rfscholte: Same here

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        super( "[3.7.0,)" );
+    }
+
+    /**
+     * Test that the --resume-from flag resolves dependencies inside the same Maven project
+     * without having them installed first.
+     * This test case uses the target/classes folder of module-a.
+     */
+    public void testShouldResolveUnpackagedArtifactFromEarlierBuild() throws Exception
```
rfscholte: method name is a bit confusing. 'Unpackaged' might suggest the artifact is unpackaged in the process. Here you want to use the outputDirectory.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
mthmulders: Oops, removed.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
mthmulders: Oops, removed.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6090CIFriendlyTest.java
```diff
@@ -19,7 +19,6 @@
  * under the License.
  */
 
-import org.apache.maven.it.Verifier;
```
rfscholte: unrelated change

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660ResumeFromTest.java
mthmulders: Agreed, updated method name as well as Javadoc.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6090CIFriendlyTest.java
mthmulders: True. We were investigating that test case and probably the IDE thought it'd be good to cleanup unused imports. Prefer to have it reverted? I think it won't hurt either....

rfscholte: It will look very weird if we go through the commit logs and see the list of files touched by this commit. It scares me if I see MNG6090 between MNG4660

mthmulders: Don't be scared, it's been reverted already.

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
rfscholte: I will update the javadoc. I've found several places that require an update

PR #60, opened 2020-05-22
[MNG-6913] Use Java 8

mthmulders:


michael-o:
I don't know whether this is a good idea. This may mean that older Maven versions cannot run ITs on Java 7 anymore.

mthmulders:
> I don't know whether this is a good idea. This may mean that older Maven versions cannot run ITs on Java 7 anymore.

Thanks for bringing that up two weeks ago, when I asked for "any objections to changing maven-integration-testing so it uses Java 8 for compilation as well?" on the dev mailing list.

michael-o:
> 
> 
> > I don't know whether this is a good idea. This may mean that older Maven versions cannot run ITs on Java 7 anymore.
> 
> Thanks for bringing that up two weeks ago, when I asked for "any objections to changing maven-integration-testing so it uses Java 8 for compilation as well?" on the dev mailing list.

Sorry, I am a very slow responder. Shall I repost on the dev list?

PR #63, opened 2020-06-04
[MNG-5868] no duplicate artifacts in attached artifacts

olamy:


olamy:
it for maven core pr https://github.com/apache/maven/pull/358

PR #70, opened 2020-08-04
[MNG-6974] fix mng5175 by waiting longer for read timeout to happen

mickroll:
spotted by @michael-o:
> the thread sleeps for 15 ms and the read timeout is set to 10 ms. I consider this is to be a racing condition

fixes MNG-6974

This PR should fix build stability problems in https://github.com/apache/maven/pull/368


mickroll:
This PR should fix build stability problems in https://github.com/apache/maven/pull/368

PR #71, opened 2020-09-09
[MNG-6981] Include child projects of projects selected with --pl

MartinKanters:
Related JIRA issue: https://issues.apache.org/jira/browse/MNG-6981.
PR for implementation is here: https://github.com/apache/maven/pull/374.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
@@ -64,7 +64,6 @@ public class MavenITmng4262MakeLikeReactorDottedPathTest
         verifier.resetStreams();
 
         verifier.assertFilePresent( "target/touch.txt" );
-        verifier.assertFileNotPresent( "../sub-a/target/touch.txt" );
     }
 
     /**
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
@@ -64,6 +64,14 @@ public class MavenITmng4262MakeLikeReactorDottedPathTest
         verifier.resetStreams();
 
         verifier.assertFilePresent( "target/touch.txt" );
+        if ( matchesVersionRange( "[,3.7.0)" ) )
+        {
+            verifier.assertFileNotPresent( "../sub-a/target/touch.txt" );
+        }
+        else if ( matchesVersionRange( "[3.7.0,)" ) )
+        {
+            verifier.assertFilePresent( "../sub-a/target/touch.txt" );
+        }
     }
 
     /**
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
@@ -29,7 +29,7 @@ public class MavenITmng6981ProjectListShouldIncludeChildrenTest
     }
 
     /**
-     * Since --pl's behavior is changed, make sure the alternative for building a pom without it's children still works.
+     * Since --pl's behavior is changed, make sure the alternative for building a pom without its children still works.
      */
     public void testFileSwitchAllowsExcludeOfChildren()
             throws Exception
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
@@ -20,7 +20,7 @@ public class MavenITmng6981ProjectListShouldIncludeChildrenTest
             throws Exception
     {
         final File testDir = ResourceExtractor.simpleExtractResources( getClass(), RESOURCE_PATH );
-        Verifier verifier = new Verifier( testDir.getAbsolutePath() );
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
 
         verifier.addCliOption( "-pl" );
         verifier.addCliOption( ":module-a" );
@@ -35,7 +35,7 @@ public class MavenITmng6981ProjectListShouldIncludeChildrenTest
             throws Exception
     {
         final File testDir = ResourceExtractor.simpleExtractResources( getClass(), RESOURCE_PATH );
-        Verifier verifier = new Verifier( testDir.getAbsolutePath() );
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
 
         verifier.addCliOption( "-f" );
         verifier.addCliOption( "module-a" );
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
@@ -1,0 +1,68 @@
+package org.apache.maven.it;
+
+import org.apache.maven.it.util.ResourceExtractor;
+
+import java.io.File;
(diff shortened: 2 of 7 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6981ProjectListShouldIncludeChildrenTest.java
```diff
@@ -0,0 +1,68 @@
...
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+
+        verifier.addCliOption( "-pl" );
+        verifier.addCliOption( ":module-a" );
+        verifier.executeGoal( "compile" );
+        verifier.verifyTextInLog( "Building module-a-1 1.0" );
+    }
+
+    /**
+     * Since --pl's behavior is changed, make sure the alternative for building a pom without it's children still works.
```
mthmulders, line 32: I think this should be in the documentation somewhere, too. Not just here in a Javadoc, but maybe in the [Guide to Working with Multiple Modules](https://maven.apache.org/guides/mini/guide-multiple-modules.html)?

MartinKanters: Good idea. I have created a [PR on maven-site](https://github.com/apache/maven-site/pull/199) about this alternative, but also some other docs that were missing. Is this what you had in mind?

I think we should extend the PR with more documentation related to the other reactor changes coming in 3.7.0.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
```diff
@@ -64,7 +64,6 @@ public void testitMakeRoot()
         verifier.resetStreams();
 
         verifier.assertFilePresent( "target/touch.txt" );
-        verifier.assertFileNotPresent( "../sub-a/target/touch.txt" );
```
michael-o, line 67: Why is this one gone?

core-it-suite/src/test/resources/mng-6981-pl-should-include-children/module-a/module-a-1/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
michael-o, line 24: Please use the canonical schema location, not the old one.

core-it-suite/src/test/resources/mng-6981-pl-should-include-children/module-a/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
michael-o, line 24: same here.

core-it-suite/src/test/resources/mng-6981-pl-should-include-children/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
michael-o, line 24: same here.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
MartinKanters: This test used to verify that if you execute `mvn <goal> -pl .` that root is built (line 66) and that the subproject is not built (line 67).
The change in MNG-6981 changed the behavior of `-pl`. First it didn't build the target project's children, now it does. 

I decided to remove the assertion on line 67, since it is in my opinion an unnecessary assertion which is not related to unit test. Also by removing the assertion we make sure that this test will work pre and post MNG-6981. Alternatively we could duplicate the test class with different version restrictions. What do you think?

michael-o: Does the test still pass with 3.6.3? The are assertions which can be scoped with version ranges. Maybe we could utilize them?

core-it-suite/src/test/resources/mng-6981-pl-should-include-children/module-a/module-a-1/pom.xml
MartinKanters: Hmm, I'm afraid I don't know enough of XML to understand what you mean. I have also checked some other POMs in the repo and they are quite similar (some are missing these three attributes completely).
Can you please give me a pointer here?

michael-o: http://maven.apache.org/xsd/maven-4.0.0.xsd

MartinKanters: Thanks! I've changed them

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4262MakeLikeReactorDottedPathTest.java
MartinKanters: I have not found assertions with version scopes, but I've found the method `matchesVersionString` which I applied. I have verified it by running Maven 3.6.3 and the MNG-6981 PR against the ITs. WDYT?

michael-o: That's exactly what I was referring to!

rfscholte: The general rule is that you should never adjust the test itself. If the behavior changes, adjust the Maven versionsRange and make a new integrationtest for it (with the new behavior)

MartinKanters: Ah, that's right, I forgot about that. Too bad we'll duplicate the tests then, but it makes sense to have the old tests 'immutable'. I took your advice, @rfscholte , this is what you meant, right?

rfscholte: Almost correct, you should only adjust the value in the constructor, this one should be

    super( "(,3.7.0)" );

MartinKanters: Oh.. my `git push` didn't get through... I've pushed the changes now. Please take another look..

mthmulders (approved):
LGTM!

rfscholte (approved):

rfscholte:
LGTM

PR #72, opened 2020-09-10
[MNG-6985] Use correct maven.multiModuleProjectDirectory in Embedded launcher

mthmulders:
The implementation is similar to what the shell scripts do.


rfscholte (approved):

mthmulders:
Merged in 2dc4f0cd.

PR #74, opened 2020-10-24
[MNG-6566] Plugins that require a certain phase should not fork goals that are already in the execution plan

MartinKanters:
Draft, work needed:
- [X] Implement fix in maven-core, see https://github.com/apache/maven/pull/397
- [ ] ~~Duplicate tests for `@Execute`'s lifecycle and goal parameters~~

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest.java
@@ -52,6 +52,9 @@ public class MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest
         verifier.verifyErrorFreeLog();
     }
 
+    /**
+     * This test uses the <pre>run</pre> goal of the test plugin which requires the 'compile' phase.
+     */
     public void testRunsCompileGoalOnceWithDirectPluginInvocation()
             throws Exception
     {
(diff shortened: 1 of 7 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest.java
```diff
@@ -0,0 +1,101 @@
...
+            throws Exception
+    {
+        File consumerDir = new File( testDir, "consumer" );
+
+        Verifier verifier = newVerifier( consumerDir.getAbsolutePath() );
+        verifier.setLogFileName( "log-phase-execution.txt" );
+        verifier.executeGoal( "compile" );
+        verifier.resetStreams();
+        verifier.verifyErrorFreeLog();
+        verifyTextNotInLog( verifier, ">>> plugin:1.0-SNAPSHOT:run (default) > compile @ consumer >>>" );
```
rfscholte, line 76: I don't think the forking is the issue, but that the same goal is being executed twice.
What I would do is  a count of all goals being executed and ensure every call is unique.

MartinKanters: Right, sounds more like what we exactly want to fix in this issue indeed. I will rework it, thanks!

mthmulders: We've updated the test to count the invocations of the M-C-P : compile goal.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest.java
```diff
@@ -0,0 +1,102 @@
...
+    extends AbstractMavenIntegrationTestCase
+{
+    private static final String RESOURCE_PATH = "/mng-6566-execute-annotation-should-not-re-execute-goals";
+    private static final String PLUGIN_KEY = "org.apache.maven.its.mng6566:plugin:1.0-SNAPSHOT";
+
+    private File testDir;
+
+    public MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest()
+    {
+        super( "[3.6.3,)" );
```
rfscholte, line 40: Don't think this is the right version

rfscholte (changes requested):
Can't you reuse https://github.com/apache/maven-integration-testing/tree/master/core-it-support/core-it-plugins/maven-it-plugin-core-stubs/maven-compiler-plugin ?

rfscholte (approved):

PR #76, opened 2020-11-09
[MNG-6754] Verify equal version timestamp in multi-module projects

mthmulders:


michael-o:
I will happily review this IT. Thank you very much.

michael-o (changes requested):
Several issues:
* Don't deploy outside of the project dir, this will cause clutter. Create `repo` in basedir
* You have written maven build timestamp, but did not use it
* If you look closely at my patch you'll notice that timestamps are in local repository AND remote repository
* You completely ignore the timestamp of the parent
* Please add the wait between all modules

What you need is maven build timestamp, 3 timestamps from local repo + 3 from remote repo. All need to be identical.

mthmulders:
> Don't deploy outside of the project dir, this will cause clutter. Create `repo` in basedir

✔️

> You have written maven build timestamp, but did not use it

I don't understand what you mean.

> If you look closely at my patch you'll notice that timestamp are in local repository AND remote repo

✔️

> You completely ignore the timestamp of the parent

✔️

>  Please add the wait between all modules

✔️

> What you need is maven build timestamp, 3 timestamps from local repo + 3 from remote repo. All need to be identical.

✔️

If you look closely at my patch, you'll notice that the version range was not correct. That has been addressed.

michael-o:
Will look up tomorrow.

michael-o:
I am going through performing some improvements.

michael-o:
I have pushed an updated branch for gitbox based on your work. Will run a few more tests. If everything is fine, this is going to be merged. Thank you!

michael-o:
I gave it more thought and think this is incomplete. The purpose of this change is not only to align last updated field, but also to make snapshot timestamp consistent throughout the reactor. This is also the narrative of the JIRA issue. This one is good for part one. I will add part two to it. Stay tuned.

michael-o:
Superseded by #77.

PR #77, opened 2020-11-11
Mng 6754 version timestamp in multimodule build

michael-o:


michael-o:
@mthmulders I have completely picked up your work and put even more validation on top of it.

mthmulders (approved):

PR #79, opened 2020-11-14
[MNG-7022] Remove o.a.m.lifecycle.mapping.Lifecycle optional mojos ba…

michael-o:
…ckward compat code

Disable test which still declares optional mojos. The test runs fine
when optional mojos are removed from components.xml.


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4338OptionalMojosTest.java
```diff
@@ -35,7 +35,7 @@
 
     public MavenITmng4338OptionalMojosTest()
     {
-        super( "[3.0,)" );
+        super( "[3.0,3.7.0)" );
```
rfscholte, line 38: can be pushed to 4.0.0-alpha-1 :)

michael-o: Done locally

PR #80, opened 2020-11-15
Add checksums to avoid warnings in ITs

michael-o:


rfscholte:
Do you expect me to verify every checksum? 
It makes sense to add them, especially when fixing https://issues.apache.org/jira/browse/MNG-5728

rfscholte (approved):

michael-o:
No, I don't. Just don't want to leave impression to add something like this via someone's back.

michael-o:
As for MNG-5768, correct. That would be a prerequisite and nice to have for 4.0.0.

PR #83, opened 2020-11-28
Fix version requirement to 4.0.0-alpha-1

michael-o:


rfscholte:
If 4.0.0 is used, are the test skipped? I updated the version in Maven to ensure alpha releases, but it that implies that we get ugly versions here, then I don't mind changing it to 4.0.0

michael-o:
This isn't a problem, when you grep through the code you will find stuff like 3.1.0-alpha-1. Since alpha comes before a GA version, e.g., 4.0.0 this has to work. I will test anyway and let you know. Our version resolution handles this.

rfscholte:
Since @mthmulders offers a brew installation of Mavens master branch, it would be nice to keep the alpha-1 in the version.

mthmulders:
> Since @mthmulders offers a brew installation of Mavens master branch, it would be nice to keep the alpha-1 in the version.

I assume you mean "so that it is very, very clear that these are unstable releases", right?

michael-o:
I do not intend to exchange 3.7.0 for 4.0.0, but for the first alpha as we have done with other tests. ITs are running locally and look good. All previous with 3.7.0 created ITs flawlessly pass with 4.0.0-SNAPSHOT with the 4.0.0-alpha-1 version.

michael-o:
Tests pass, none of the new tests are skips. From my POV this is good to merge.

rfscholte:
@mthmulders exactly. Not sure if 'unstable' is the right word. It should make clear there might be edge cases that we haven't thought of yet and that we expect feedback.

mthmulders:
> @mthmulders exactly. Not sure if 'unstable' is the right word. It should make clear there might be edge cases that we haven't thought of yet and that we expect feedback.

Perfect. Just wanted make sure that those builds should not pose technical limitations on what we think is the best solution to any given problem. They're my personal pet project, not an official community offering, as far as I'm concerned.

PR #84, opened 2020-11-30
[MNG-5728] Switch the default checksum policy from "warn" to "fail"

michael-o:
To pass all tess on Windows you need a fresh clone, not a checkout!


rfscholte (approved):
Great work! It must have taken quite some time to collect all checksums, but with this we can safely turn on strict checksums by default

michael-o:
> 
> 
> Great work! It must have taken quite some time to collect all checksums, but with this we can safely turn on strict checksums by default

Yes, but 99% of the checksums have been done by:
```
find .  -name repo\*  -type d -exec tree -ifF {} + | grep -v -E -e '/$' -e 'repo(-[a-z0-9])?$'   -e .md5 -e .sha1 | xargs -I^ sh -c 'sha1 -q ^ > ^.sha1 && truncate -s -1 ^.sha1'
```

Unix magic, I am a lazy guy.

PR #87, opened 2020-12-09
[MNG-7051] Test activation of non-existing profiles

mthmulders:
Integration tests for [MNG-7051](https://issues.apache.org/jira/browse/MNG-7051). Some are duplicated (a pre-4.0.0 and a post-4.0.0 version) because the behaviour has changed in a breaking way.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
@@ -106,6 +106,7 @@ public class IntegrationTestSuite
         // Tests that don't run stable and need to be fixed
         // -------------------------------------------------------------------------------------------------------------
         // suite.addTestSuite( MavenIT0108SnapshotUpdateTest.class ); -- MNG-3137
+        suite.addTestSuite( MavenITmng7051OptionalProfileActivationTest.class );
         suite.addTestSuite( MavenITmng6957BuildConsumer.class );
         suite.addTestSuite( MavenITmng7045DropUselessAndOutdatedCdiApiTest.class );
         suite.addTestSuite( MavenITmng6566ExecuteAnnotationShouldNotReExecuteGoalsTest.class );
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2309ProfileInjectionOrderTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2309ProfileInjectionOrderTest.java
@@ -22,7 +22,6 @@ package org.apache.maven.it;
 import java.io.File;
 import java.util.Properties;
 
-import org.apache.maven.it.Verifier;
 import org.apache.maven.it.util.ResourceExtractor;
 
 /**
@@ -52,10 +51,19 @@ public class MavenITmng2309ProfileInjectionOrderTest
         verifier.setAutoclean( false );
         verifier.addCliOption( "--settings" );
         verifier.addCliOption( "settings.xml" );
-        verifier.addCliOption( "-P"
-            + "pom-a,pom-b,pom-e,pom-c,pom-d"
-            + ",profiles-a,profiles-b,profiles-e,profiles-c,profiles-d"
-            + ",settings-a,settings-b,settings-e,settings-c,settings-d" );
+        if ( matchesVersionRange( "[4.0.0-alpha-1,)" ) )
+        {
+            verifier.addCliOption( "-P"
+                    + "pom-a,pom-b,pom-e,pom-c,pom-d"
+                    + ",settings-a,settings-b,settings-e,settings-c,settings-d" );
+        }
+        else
+        {
+            verifier.addCliOption( "-P"
+                    + "pom-a,pom-b,pom-e,pom-c,pom-d"
+                    + ",profiles-a,profiles-b,profiles-e,profiles-c,profiles-d"
+                    + ",settings-a,settings-b,settings-e,settings-c,settings-d" );
+        }
         verifier.executeGoal( "validate" );
         verifier.verifyErrorFreeLog();
         verifier.resetStreams();
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3641ProfileActivationWarningTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3641ProfileActivationWarningTest.java
@@ -35,7 +35,7 @@ public class MavenITmng3641ProfileActivationWarningTest
 
     public MavenITmng3641ProfileActivationWarningTest()
     {
-        super( "[2.0.11,2.1.0-M1),[2.1.0,)" ); // only test in 2.0.11+, 2.1.0+
+        super( "[2.0.11,2.1.0-M1),[2.1.0,4.0.0-alpha-1)" ); // only test in 2.0.11+, 2.1.0+
     }
 
     public void testitMNG3641()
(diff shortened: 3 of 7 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7051NonExistingProfileActivationTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        try
+        {
+            verifier.executeGoal( "validate" );
+            fail( "Activated a non-existing profile without ? prefix should break the build, but it didn't." );
+        }
+        catch ( VerificationException ve )
+        {
+            // Inspect the reason why the build broke.
+            verifier.verifyTextInLog( "The requested profile(s) \"non-existing-profile\" could not be activated or de-activated because they do not exist." );
```
michael-o, line 57: Make sure that you verify a warning with the `[WARNING]` prefix, some other ITs do this too.

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
```diff
@@ -106,6 +106,10 @@ public static Test suite()
         // Tests that don't run stable and need to be fixed
         // -------------------------------------------------------------------------------------------------------------
         // suite.addTestSuite( MavenIT0108SnapshotUpdateTest.class ); -- MNG-3137
+        suite.addTestSuite( MavenITmng4106InterpolationUsesDominantProfile400Test.class );
```
michael-o, line 109: I am a bit confused by this. Some of the are already there, why they pop up here now and why have some disabled?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
MartinKanters: The original tests fail starting with the introduced optional behavior in Maven 4, that's why we copied the original testcase and slightly adjusted them (and appended 400 to the class name, meaning version 4.0.0).

We have done this as well for the MavenITmng4262MakeLikeReactorDottedPathTest testcase in another feature on @rfscholte's advice.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7051NonExistingProfileActivationTest.java
MartinKanters: Nice find. I have added the severity prefixes.

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
michael-o: Hmm, may ITs have an inline if for those cases. Why not use that style? It reduces the amount of code we have. Example: https://github.com/apache/maven-integration-testing/pull/96/files#diff-0323fd1cd7d6126ff8efaf53b1771bb88deb78c458141d38695979480c6710e5R58

MartinKanters: My opinion is that those ifs can be useful in the tests when only a single thing has to be changed per version. 
In the classes that we have duplicated we noticed that we would have to change the `.addCliOption` (which activates the profiles) everywhere and that these tests already contained big `if (mvnVersion < 3) `-branches which tested support for profiles.xml.

So in this case my preference is to duplicate the test class to keep the tests readable and immutable (which was @rfscholte 's advice).

It does increase the amount of code + tests indeed, but I don't think that outweighs the benefits listed above. 
What do you think?

michael-o: My general rule of thumb is if the test has a different structure or requires 50+% in code modification then it needs to be copied. I need to take a closer look how the diff actually looks like.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2309ProfileInjectionOrder400Test.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.File;
+import java.util.Properties;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-2309">MNG-2309</a>.
+ *
+ * @author Benjamin Bentmann
+ *
+ */
+public class MavenITmng2309ProfileInjectionOrder400Test
```
michael-o, line 33: Call me stupid, but this code looks almost identical to the previous one and the test resources are identical except the `profiles.xml`...

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3732ActiveProfiles400Test.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-3732">MNG-3732</a>.
+ *
+ * @author Benjamin Bentmann
+ */
+public class MavenITmng3732ActiveProfiles400Test
```
michael-o, line 36: This is virtually identical too.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4106InterpolationUsesDominantProfile400Test.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.File;
+import java.util.Properties;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-4106">MNG-4106</a>.
+ *
+ * @author Benjamin Bentmann
+ *
+ */
+public class MavenITmng4106InterpolationUsesDominantProfile400Test
```
michael-o, line 33: Same here.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7051NonExistingProfileActivationTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.apache.maven.it.util.ResourceExtractor;
+
+import java.io.File;
+import java.io.IOException;
+
+public class MavenITmng7051NonExistingProfileActivationTest
+        extends AbstractMavenIntegrationTestCase
+{
+    private static final String PROJECT_PATH = "/mng-7051-non-existing-profile-activation";
```
michael-o, line 30: the trailing summary isn't true since you have test for an existing profile in the last method.

michael-o (changes requested):
I wen't through all tests. While the actual test if fine, I am absolutely not convinced in the test copies because they are 95% identical. There are now two ways to solve it:

* Modify the existing ones like I proposed (did already in other commits)
* If you still think they need to be separate, move/copy them in the the new test class along with resources then you have a compled test set.

MartinKanters:
> I wen't through all tests. While the actual test if fine, I am absolutely not convinced in the test copies because they are 95% identical. There are now two ways to solve it:
> 
> * Modify the existing ones like I proposed (did already in other commits)
> * If you still think they need to be separate, move/copy them in the the new test class along with resources then you have a compled test set.

Alright, I've merged them in the original tests. Still not a fan, but it turned out better than I imagined.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7051NonExistingProfileActivationTest.java
MartinKanters: Nice find. Fixed this as well. (classname + directory name)

MartinKanters:
@michael-o I added tests for the `!?` scenario.

michael-o (approved):
Looks good now. Please squash and merge.

PR #88, opened 2020-12-14
Drop CDI API from Maven

huembert:
just some initial code


michael-o:
Will run tests and let you know.

michael-o:
Going through now...

PR #89, opened 2020-12-22
Fix mng-3259 test on JDK 16

famod:
Popped up here: https://github.com/apache/maven/pull/413#issuecomment-748678853

/cc @michael-o


michael-o:
Does this work on a empty repo? I most cases deps which are not in bootstrap aren't found.

famod:
@michael-o Well, after running the tests `xstream-1.4.15.jar` is present in `maven-integration-testing/repo` but _not_ in my regular local repo. So LGTM.

PS: I'm using `run-its.sh`.

PR #90, opened 2020-12-23
[MNG-4660] Increase usefulness of logging

mthmulders:
When the user resumes a build, they could see a message "Packaged artifact is not up-to-date compared to the build output directory". This could even be logged multiple times for one build. It was unclear why it was logged and what a user could do to address it. This test accompanies https://github.com/apache/maven/pull/414.


mthmulders:
Merged in acbb518ac.

PR #91, opened 2020-12-23
[MNG-4660] Increase usefulness of logging (2)

mthmulders:
When the user resumes a build, they could see a message "Packaged artifact is not up-to-date compared to the build output directory". This could even be logged multiple times for one build. It was unclear why it was logged and what a user could do to address it. This test accompanies apache/maven#416.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
@@ -23,12 +23,18 @@ import org.apache.maven.it.util.ResourceExtractor;
 import org.apache.maven.shared.utils.io.FileUtils;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.FileVisitor;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.BasicFileAttributes;
 
 import static org.junit.Assert.assertThat;
 import static org.hamcrest.Matchers.greaterThan;
+
+import static java.nio.file.FileVisitResult.CONTINUE;
 
 /**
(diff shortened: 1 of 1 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
```diff
@@ -112,7 +112,9 @@ public void testShouldWarnWhenPackagedArtifactIsOutdated() throws Exception
         verifier3.verifyErrorFreeLog();
         try
         {
-            verifier3.verifyTextInLog( "File 'module-a/target/classes/example.properties' is more recent than the packaged artifact for 'module-a'; using 'module-a/target/classes' instead" );
+            Path path = Paths.get( "module-a", "target", "classes", "example.properties" );
+            verifier3.verifyTextInLog( "File '" + path.toString() + "' is more recent than the packaged artifact for 'module-a'; "
+                    + "using 'module-a/target/classes' instead" );
```
michael-o, line 117: Is this line portable too?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4660OutdatedPackagedArtifact.java
```diff
@@ -112,7 +112,9 @@ public void testShouldWarnWhenPackagedArtifactIsOutdated() throws Exception
         verifier3.verifyErrorFreeLog();
         try
         {
-            verifier3.verifyTextInLog( "File 'module-a/target/classes/example.properties' is more recent than the packaged artifact for 'module-a'; using 'module-a/target/classes' instead" );
+            Path path = Paths.get( "module-a", "target", "classes", "example.properties" );
+            verifier3.verifyTextInLog( "File '" + path.toString() + "' is more recent than the packaged artifact for 'module-a'; "
```
michael-o, line 116: the #toString() is redundant. It will be invoked automatically.

PR #93, opened 2020-12-26
[MNG-5760] Add `-r/--resume` to automatically resume from the last fa…

michael-o:
…ilure point

Use a fresh, preconfigured verifier and do not clear CLI options because
on clear all preconfigurations, especially global settings are gone.

This issue can be observed when Maven is ran with MNG-4645 where
preconfigured Maven Central is lost in global settings.


mthmulders (approved):
LGTM

MartinKanters (approved):
LGTM, nice find

PR #95, opened 2020-12-26
[MNG-4991] LegacyRepositorySystem#injectProxy(repositories, proxies) …

michael-o:
…doesn't evaluate non-proxy hosts

Remove empty override of 'it-defaults' profile because it breaks MNG-4645. It
still reliably on both cases because:
before MNG-4645) It relied on the fact that the override removes the global
settings Maven Central definition of 'file:target/null' while Maven internally
still added canonical Maven Central back.
with MNG-4645) The empty profile override completely removes Maven Central
definition which means that the mirror won't proxy a non-existing repository.

This closes #95


michael-o:
@gnodet Is this and its ticket obsolete with the move of Central out of core?

PR #96, opened 2020-12-26
[MNG-4963] Parent POM not downloaded when settings define global mirr…

michael-o:
…or and one snapshot repo but no other release repository

This IT was written with implicit black magic in mind. A mirror simply exchanges
a URL of repo with another one while retaining the entire config for
releases/snapshots. Logically this IT should have failed because there is no
release repo. But it works because Maven adds Maven Central internally which is
a release repo. For MNG-4645 one has to add a release repo in the settings to
mimic that behavior.

This closes #96


michael-o:
@gnodet Is this and its ticket obsolete with the move of Central out of core?

PR #99, opened 2021-02-12
[MNG-7099] Add integration test

gnodet:


gnodet:
Duplicate, already fixed in maven 4

PR #100, opened 2021-02-13
[MNG-6511] - Optional project selection

MartinKanters:
Relevant JIRA issue: https://issues.apache.org/jira/browse/MNG-6511

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
@@ -44,6 +44,8 @@ public class MavenITmng6511OptionalProjectSelectionTest extends AbstractMavenInt
 
     public void testSelectExistingOptionalProfile() throws VerificationException
     {
+        newVerifier( testDir.getAbsolutePath() ).executeGoal( "clean" );
+
         Verifier verifier = newVerifier( testDir.getAbsolutePath() );
         verifier.setLogFileName( "log-select-existing.txt" );
         verifier.addCliOption( "-pl ?existing-module" );
@@ -54,6 +56,8 @@ public class MavenITmng6511OptionalProjectSelectionTest extends AbstractMavenInt
 
     public void testSelectExistingOptionalProfileByArtifactId() throws VerificationException
     {
+        newVerifier( testDir.getAbsolutePath() ).executeGoal( "clean" );
+
         Verifier verifier = newVerifier( testDir.getAbsolutePath() );
         verifier.setLogFileName( "log-select-existing-artifact-id.txt" );
         verifier.addCliOption( "-pl ?:existing-module" );
@@ -64,6 +68,8 @@ public class MavenITmng6511OptionalProjectSelectionTest extends AbstractMavenInt
 
     public void testSelectNonExistingOptionalProfile() throws VerificationException
     {
+        newVerifier( testDir.getAbsolutePath() ).executeGoal( "clean" );
+
         Verifier verifier = newVerifier( testDir.getAbsolutePath() );
         verifier.setLogFileName( "log-select-non-existing.txt" );
         verifier.addCliOption( "-pl ?non-existing-module" );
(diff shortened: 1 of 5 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void testSelectNonExistingOptionalProfile() throws VerificationException
+    {
+        newVerifier( testDir.getAbsolutePath() ).executeGoal( "clean" );
+
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+        verifier.setLogFileName( "log-select-non-existing.txt" );
+        verifier.addCliOption( "-pl ?non-existing-module" );
+        verifier.executeGoal( "validate" );
+        verifier.verifyErrorFreeLog();
+        verifier.assertFilePresent( "existing-module/target/touch.txt" ); // existing-module should have been built.
```
michael-o, line 78: How does this line relate to `non-exsisting-module`?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void testDeselectNonExistingOptionalProfile() throws VerificationException
+    {
+        newVerifier( testDir.getAbsolutePath() ).executeGoal( "clean" );
+
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+        verifier.setLogFileName( "log-deselect-non-existing.txt" );
+        verifier.addCliOption( "-pl !?non-existing-module" );
+        verifier.executeGoal( "validate" );
+        verifier.verifyErrorFreeLog();
+        verifier.assertFilePresent( "existing-module/target/touch.txt" ); // existing-module should have been built.
```
michael-o, line 102: Same here.

core-it-suite/src/test/resources/mng-6511-optional-project-selection/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  </properties>
+
+  <modules>
+    <module>existing-module</module>
+  </modules>
+
+  <build>
+    <pluginManagement>
+      <!-- maven-resources-plugin, maven-compiler-plugin -->
+      <plugins>
```
michael-o, line 51: Did you check whether these are downloaded via bootstrap IT? If not, use the same versions as bundled with Maven 4.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
MartinKanters: Well, our idea was the following, if someone tries to select a module optionally and it does not exist, the build should just build everything. In other words `mvn validate -pl ?non-existing` has the same outcome as `mvn validate`.

core-it-suite/src/test/resources/mng-6511-optional-project-selection/pom.xml
MartinKanters: Yep, [they are](https://github.com/apache/maven-integration-testing/blob/master/core-it-suite/src/test/resources/bootstrap/pom.xml#L61-L80) :)

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6511OptionalProjectSelectionTest.java
michael-o: Alright, I have noticed that in other tests too. Maybe this special case should be documented where we have profile selection in maven-site. By defiinition, non-existing will cause the entire build. Take it or leave it.

MartinKanters: Yeah, sounds like a good idea! Updating the docs in maven-site is something we shouldn't forget, and especially this case.

michael-o: Please create a MNGSITE ticket for this.

mthmulders: See [MNGSITE-444](https://issues.apache.org/jira/browse/MNGSITE-444).

michael-o (approved):

PR #104, opened 2021-03-22
[MNG-7112] --projects flag should honor --non-recursive flag

MartinKanters:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7112ProjectsWithNonRecursiveTest.java
```diff
@@ -0,0 +1,78 @@
...
+        verifier.addCliOption( "!:aggregator-a,!:aggregator-b" );
+        verifier.addCliOption( "-N" );
+        verifier.setLogFileName( "excluded-non-recursive.txt" );
+        verifier.executeGoal( "validate" );
+
+        verifier.assertFilePresent( "target/touch.txt" );
+        verifier.assertFileNotPresent( "aggregator-a/target/touch.txt" );
+        verifier.assertFilePresent( "aggregator-a/module-a/target/touch.txt" );
+        verifier.assertFileNotPresent( "aggregator-b/target/touch.txt" );
+        verifier.assertFilePresent( "aggregator-b/module-b/target/touch.txt" );
```
michael-o, line 76: Can you explain this behavior? Disabling aggregators, but children are enabled with non-recursive flag?

MartinKanters: By default in the build all modules would be built. But we exclude two aggregators. Recently, we have made aggregator selection and exclusion recursive by default. In 3.6.3 this is non-recursive. So to get the old (non-recursive) behavior back, you can make use of -pl + -N to just exclude the two aggregator projects.

michael-o (approved):

PR #107, opened 2021-04-19
[GitHub CI] Use the loopback address for connecting to an in-process web server

mthmulders:
This may resolve the connection issues we sometimes see in GitHub
Actions, where the Maven process that is being tested cannot connect
to the Jetty server that was just started.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenIT0146InstallerSnapshotNaming.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenIT0146InstallerSnapshotNaming.java
@@ -86,7 +86,7 @@ public class MavenIT0146InstallerSnapshotNaming
         Verifier verifier = newVerifier( testDir.getAbsolutePath() );
 
         Properties properties = verifier.newDefaultFilterProperties();
-        properties.setProperty( "@host@", InetAddress.getLocalHost().getCanonicalHostName() );
+        properties.setProperty( "@host@", InetAddress.getLoopbackAddress().getCanonicalHostName() );
         properties.setProperty( "@port@", Integer.toString( port ) );
 
         verifier.filterFile( "settings-template.xml", "settings.xml", "UTF-8", properties );
@@ -125,7 +125,7 @@ public class MavenIT0146InstallerSnapshotNaming
         verifier = newVerifier( testDir.getAbsolutePath() );
 
         Properties properties = verifier.newDefaultFilterProperties();
-        properties.setProperty( "@host@", InetAddress.getLocalHost().getCanonicalHostName() );
+        properties.setProperty( "@host@", InetAddress.getLoopbackAddress().getCanonicalHostName() );
         properties.setProperty( "@port@", Integer.toString( port ) );
 
         verifier.filterFile( "settings-template.xml", "settings.xml", "UTF-8", properties );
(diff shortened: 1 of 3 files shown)


michael-o (commented):
Isn't `getLoopbackAddress()` suficient?

mthmulders:
> Isn't `getLoopbackAddress()` suficient?

That returns an `InetAddress`, but `setProperty` expects a `String` as the second argument...

michael-o:
> 
> 
> > Isn't `getLoopbackAddress()` suficient?
> 
> That returns an `InetAddress`, but `setProperty` expects a `String` as the second argument...

Ah, thanks!

michael-o (approved):

michael-o (commented):
Please note that some ITs explicitly require a non-localhost hostname to perform property. Did you verify that?

slachiewicz (approved):

mthmulders:
> Please note that some ITs explicitly require a non-localhost hostname to perform property. Did you verify that?

I didn't know that, so thanks for asking. I checked all three cases, and can't find any evidence in the code or in the JIRA tickets that it's crucially important to connect to a **remote** host. So in all honesty, I don't know...

michael-o:
> 
> 
> > Please note that some ITs explicitly require a non-localhost hostname to perform property. Did you verify that?
> 
> I didn't know that, so thanks for asking. I checked all three cases, and can't find any evidence in the code or in the JIRA tickets that it's crucially important to connect to a **remote** host. So in all honesty, I don't know...

When the name says `MavenITmng4991NonProxyHostsTest.java` one should be vigilant.

mthmulders:
> When the name says `MavenITmng4991NonProxyHostsTest.java` one should be vigilant.

I appreciate your vigilance. This is the relevant part of the settings-template.xml for MNG-4991:

```xml
<mirrors>
  <mirror>
    <id>central</id>
    <!-- NOTE: We need to try and use the proper host name/ip as Java generally ignores proxies for "localhost" -->
    <url>http://@localhost@:@port@/</url>
    <mirrorOf>central</mirrorOf>
  </mirror>
</mirrors>
<proxies>
  <proxy>
    <active>true</active>
    <protocol>http</protocol>
    <host>localhost</host>
    <port>@proxyPort@</port>
    <nonProxyHosts>@localhost@</nonProxyHosts>
  </proxy>
</proxies>
```

Quoting from [MNG-4991](https://issues.apache.org/jira/browse/MNG-4991):

> each repository will be configured with a proxy, even if the related host is excluded via the `<nonProxyHosts />` tag in the settings.xml.

This suggests that the test wants to prove that Maven will **not** use the proxy. Next, looking at the Java code, I see:
- the test spins up a proxy (that will in fact not proxy anything, as the `DefaultHandler` always returns 404)
- the test spins up another Jetty instance, that will return an artifact
- the test builds a small Maven project and verifies the artifact is indeed resolved. If it is resolved, that would be because the proxy is bypassed, QED. If Maven would've used the proxy, it would've seen HTTP 404 and the artifact would not have been resolved.

But with all this, I still don't know if it matters where the server and the repository live. That tiny line of comment suggests

> Java generally ignores proxies for "localhost"

That's something I've never heard of before. It may be true, I don't know. But the "generally" is not quite convincing me here. The [Java Networking and Proxies](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) doesn't support that statement - in fact, it even gives an example with `-Dhttp.nonProxyHosts="localhost|host.example.com"`. Why would they do that if Java indeed ignores proxies for localhost?

So, with all that vigilance, I still don't know whether it matters or not...

michael-o:
If you can wait a few days, I will have a look at the test.

mthmulders:
> If you can wait a few days, I will have a look at the test.

I'd appreciate it. Those actions aren't our final quality gate, and they're already red for some time I believe. Those extra days won't hurt - modifying a test and unknowingly making it test the opposite would definitely hurt.

michael-o:
Looking into this today...

michael-o:
Tested, I think this change is fine.

mthmulders:
Thank you for your efforts, @michael-o.

PR #108, opened 2021-07-19
[MNG-6271] Add validationRule for expressions in repositoryUrl

rfscholte:


core-it-suite/src/test/resources/mng-6759-transitive-dependency-repositories/module2/pom.xml
```diff
@@ -11,7 +11,7 @@
     <repositories>
         <repository>
             <id>customRepo</id>
-            <url>${custom.repo.uri}</url>
+            <url>file:///${basedir}/../dependency-in-custom-repo/target/repo</url>
```
michael-o, line 14: Isn't this deprecated for `{project.basedir}`?

rfscholte: According to https://maven.apache.org/ref/current/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html it is still valid

michael-o: My bad, it is indeed still valid...

michael-o (changes requested):
Plaese also change the commit message here..

rfscholte:
Will adjust the commit message after the squash and merge.

PR #114, opened 2021-09-28
Fix IT0138: the m-plugin-p mojo is removed from maven-plugin lifecycle

cstamas:
The extra mojo is not needed anymore, as Resolver handles the group
level metadata now.

Related to PR https://github.com/apache/maven/pull/555


cstamas:
Lifecycle in related PR went back to "no change" so this PR is obsolete, dropping it

PR #116, opened 2021-10-01
Fix build consumer tests on OSX

gnodet:
The tests always fail for me.  This may be caused by my git config or OSX.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6656BuildConsumer.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6656BuildConsumer.java
@@ -76,15 +76,19 @@ public class MavenITmng6656BuildConsumer
 
         String content;
         content = FileUtils.fileRead( new File( testDir, "expected/parent.pom") );
+        content = content.replaceAll( "\r\n", "\n" ).replaceAll( "\n", System.lineSeparator() );
         verifier.assertArtifactContents( "org.sonatype.mavenbook.multi", "parent", "0.9-MNG6656-SNAPSHOT", "pom", content );
 
         content = FileUtils.fileRead( new File( testDir, "expected/simple-parent.pom") );
+        content = content.replaceAll( "\r\n", "\n" ).replaceAll( "\n", System.lineSeparator() );
         verifier.assertArtifactContents( "org.sonatype.mavenbook.multi", "simple-parent", "0.9-MNG6656-SNAPSHOT", "pom", content );
 
         content = FileUtils.fileRead( new File( testDir, "expected/simple-weather.pom") );
+        content = content.replaceAll( "\r\n", "\n" ).replaceAll( "\n", System.lineSeparator() );
         verifier.assertArtifactContents( "org.sonatype.mavenbook.multi", "simple-weather", "0.9-MNG6656-SNAPSHOT", "pom", content );
 
         content = FileUtils.fileRead( new File( testDir, "expected/simple-webapp.pom") );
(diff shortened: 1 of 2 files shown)


michael-o:
I am a bit confused, this should fail on any POSIX like OS, shouldn't it?

gnodet:
> I am a bit confused, this should fail on any POSIX like OS, shouldn't it?

I think this is mainly a problem with the git config on POSIX systems.  Depending on your `auto.crlf` git attribute configuration, the file on disk can contain `\r\n` or `\n` line endings.  The generator always generates system specific line endings, but if the git local repo is not configured correctly, the test may end up comparing different kind of line endings and will fail.

This may also be the case that just a few files have problems with line endings:
```
➜  maven-integration-testing git:(master) gst
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
➜  maven-integration-testing git:(master) git add --renormalize .
➜  maven-integration-testing git:(master) ✗ gst
On branch master
Your branch is up to date with 'origin/master'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   core-it-suite/src/test/resources/it0146/project/pom.xml
	modified:   core-it-suite/src/test/resources/mng-4148/settings-template.xml
	modified:   core-it-suite/src/test/resources/mng-5898/ear/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/ejbs/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/primary-source/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/projects/logging/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/projects/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/servlets/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/servlets/servlet/pom.xml
	modified:   core-it-suite/src/test/resources/mng-5898/servlets/servlet/src/main/webapp/index.jsp

➜  maven-integration-testing git:(master) ✗ 
```

I think one of the problem comes from the `run-its.sh` script which fails for me with 
```
zsh: ./run-its.sh: bad interpreter: /bin/sh^M: no such file or directory
```
which lead me to use
```
git config auto.crlf true
```
and ending up with the failing tests.

An alternative to this commit would be the following to commit the few files mentioned above to renormalize them.  That could be a better option.

cstamas:
IMHO, if this is needed to _compare_ generated (by test) and expected (checked in git) POMs, then your environment is wrong. Actually, IMHO, the test itself is wrong, by comparing POMs as "string blobs". It would be better to load POMs (both, generated and expected) as List<String> (as line-ending would be omitted then) and just compare the two lists, as this test on Win plaf and on Unix plaf will generate different line-endings, while (depending on your git config) it may not be aligned with checked out POMs.

michael-o:
This should be default on any dev machine:
```git config auto.crlf true```

**Comment by mthmulders** [comment]
I just ran all IT's on my own Macbook, and they all pass on `master`. My Git config:

```console
$ git config -l | grep crlf 
core.autocrlf=input
```

cstamas:
Same here
```
cstamas@Urnebes ~/Worx/apache-maven/maven  (mojo-logging)$ git config -l | grep crlf
core.autocrlf=input
cstamas@Urnebes ~/Worx/apache-maven/maven  (mojo-logging)$ 
```

michael-o (commented):
I prefer the solution in #117 instead of this symptom fix.

gnodet:
It looks like both #117 and #118 are worthwhile to me, I'll close this one in favour of those twos then.

PR #118, opened 2021-10-04
Fix for MavenITmng6656BuildConsumer

cstamas:
Instead to compare POMs as "string blobs" (so along with all line-ending),
it should go for pure content instead, as it makes IT more
resilient to environment related issues.


michael-o:
This is truly system agnostic.

PR #119, opened 2021-10-05
Fix pom content comparison to avoid line ending problems

gnodet:


gnodet:
Supersedes https://github.com/apache/maven-integration-testing/pull/118 which is missing one test to fix.

cstamas (approved):

PR #120, opened 2021-10-05
[MNG-6762] Test whether a settings.xml file can be referenced relatively to the .mvn directory instead of working directory

MartinKanters:
Before, when a settings.xml file was referenced in maven.config, it was resolved against the working directory, which would break when maven is invoked in a submodule. This PR changes it to resolve against the multi module root.

Maven-core PR: https://github.com/apache/maven/pull/566


michael-o:
Can this be closed as well?

PR #123, opened 2021-11-03
[MNG-7217] Proper use of CLI

mthmulders:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6562WarnDefaultBindings.java
```diff
@@ -43,7 +43,8 @@ public void testItShouldNotWarn()
         verifier.setAutoclean( false );
         verifier.setLogFileName( phase + ".txt" );
         verifier.setForkJvm( true ); // required due to --fos-on-severity
```
michael-o, line 45: Can you change that to `fail-on-security`? It seems wrong here.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6562WarnDefaultBindings.java
mthmulders: I've taken the liberty to go for `fail-on-severity` ;-)

michael-o (approved):

PR #133, opened 2022-01-04
[MNG-6326] Make the build fail if core extensions can not be loaded

gnodet:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6326CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,56 @@
...
+ * under the License.
+ */
+
+import java.io.File;
+
+import org.apache.maven.it.util.ResourceExtractor;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-6326">MNG-6326</a>:
+ * check that Maven fails if it can not load core extensions contributed by <code>.mvn/extensions.xml</code>.
```
michael-o, line 28: cannot

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6326CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,56 @@
...
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+        try
+        {
+            verifier.executeGoal( "validate" );
+            fail( "should have failed ");
+        }
+        catch ( VerificationException e )
+        {
+            verifier.verifyTextInLog( "[ERROR] Error executing Maven." );
+            verifier.verifyTextInLog( "Plugin org.apache.maven.its.it-core-extensions:maven-it-unknown-extensions:0.1 or one of its dependencies could not be resolved: org.apache.maven.its.it-core-extensions:maven-it-unknown-extensions:jar:0.1 was not found in" );
```
michael-o, line 52: That is really weird because we aren't loading a plugin, but an extension.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6326CoreExtensionsNotFoundTest.java
gnodet: I think the behavior comes from https://github.com/apache/maven/blob/d92508179410897404bcc7e826bb0877be2d08b8/maven-embedder/src/main/java/org/apache/maven/cli/internal/BootstrapCoreExtensionManager.java#L133-L139

michael-o: WTF!

gnodet: I'll change the exception message.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6326CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-6326">MNG-6326</a>:
+ * check that Maven fails if it cannot load core extensions contributed by <code>.mvn/extensions.xml</code>.
+ */
+public class MavenITmng6326CoreExtensionsNotFoundTest
+    extends AbstractMavenIntegrationTestCase
+{
+    public MavenITmng6326CoreExtensionsNotFoundTest()
+    {
+        super( "(3.2.5,)" );
```
michael-o, line 39: 3.8.5?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6326CoreExtensionsNotFoundTest.java
gnodet: Fixed

PR #135, opened 2022-01-14
[MNG-7390] Testing whether modules can be selected outside the cwd.

MartinKanters:
Maven core PR: https://github.com/apache/maven/pull/659
Jira issue: https://issues.apache.org/jira/browse/MNG-7390

### DIFF
--- a/core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
+++ b/core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>module-a</artifactId>
+
+  <parent>
+    <groupId>org.apache.maven.its.mng7390</groupId>
+    <artifactId>parent</artifactId>
+    <version>1.0</version>
+  </parent>
+
+</project>
--- a/core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-b/pom.xml
+++ b/core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-b/pom.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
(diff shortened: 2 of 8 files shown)


core-it-suite/src/test/resources/mng-7390-pl-outside-pwd/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <packaging>pom</packaging>
+
+    <name>Maven Integration Test :: MNG-7390</name>
+    <description>
+        This test suite tests whether other modules in the same multi module project can be selected when invoking Maven from a submodule.
+    </description>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
mthmulders, line 40: Not 100% sure here... Aren't the IT's supposed to run with Java 7, too? Or can we be sure it'll be Java 8 since this test is targeted for 4.0.0-alpha-1?

mthmulders (commented):
Overall, it looks good to me. Left a small question.

core-it-suite/src/test/resources/mng-7390-pl-outside-pwd/pom.xml
michael-o: Yes, they need to be 7. Maybe this will be skipped. I have fixed similar some time ago. Check the log.

MartinKanters: Good catch. I copied the project structure over from another IT, so I checked whether there are more 1.8 tests, and there are. I'll provide a PR for fixing those soon. This one has been fixed at least. Ready to be reviewed again.

mthmulders (approved):

michael-o (changes requested):
The title is wrong so is the test class name:
* `pwd` = command to print current working dir
* `cwd` = denotes the current working dir

See: https://pubs.opengroup.org/onlinepubs/007904975/functions/getcwd.html

Please rename

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
michael-o, line 24: This location is obsolete

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <artifactId>module-a</artifactId>
+
+    <parent>
+        <groupId>org.apache.maven.its.mng7390</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0</version>
+    </parent>
+
+</project>
```
michael-o, line 36: Our default indent for XML is two spaces

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-b/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+
+    <artifactId>module-b</artifactId>
+
+    <parent>
+        <groupId>org.apache.maven.its.mng7390</groupId>
+        <artifactId>parent</artifactId>
+        <version>1.0</version>
+    </parent>
+
+</project>
```
michael-o, line 36: Same here

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+                        <goals>
+                            <goal>reset</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
```
michael-o, line 71: Same here

michael-o (changes requested):
So this will fail of there is no `.mvn/`? If so, this likely needs to be covered in the IT as well?

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
MartinKanters: Removed it

core-it-suite/src/test/resources/mng-7390-pl-outside-cwd/module-a/pom.xml
MartinKanters: Didn't know. I'm afraid most of the IT projects are in either 4, 2 or some hybrid :( Anyway, made mine 2 spaces. 
Not sure if we want to make all the XML files 2 spaces, that would be a large commit and not sure if it would do good for the git history.

PR #136, opened 2022-02-09
MNG-7395 - add a test with properties in extensions.xml

laeubi:
This is a testcase for

- https://github.com/apache/maven/pull/673
- https://github.com/apache/maven/pull/665


michael-o:
I would recommend to test with {{maven.config}} as well.

laeubi:
> I would recommend to test with {{maven.config}} as well.

done, I just don't know how this change could be reflected in the PR that changes the code.

PR #137, opened 2022-02-11
Set Maven compiler and release versions to 7 for integration test projects

MartinKanters:
See discussion here: https://github.com/apache/maven-integration-testing/pull/135#discussion_r787104645.

### DIFF
--- a/core-it-suite/src/test/resources/mng-7045/pom.xml
+++ b/core-it-suite/src/test/resources/mng-7045/pom.xml
@@ -25,8 +25,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.8.1</version>
         <configuration>
-          <source>8</source>
-          <target>8</target>
+          <source>1.7</source>
+          <target>1.7</target>
           <encoding>UTF-8</encoding>
         </configuration>
       </plugin>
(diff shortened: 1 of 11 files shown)


michael-o (approved):
Shit, this should have never happened. Thank you for being vigilant!

mthmulders (approved):
```sh
$ grep -R "source>1.8" core-it-suite/src/test/resources | wc -l
9
```
I guess you had all occurrences :-)

michael-o:
Check with >8 too

MartinKanters:
@mthmulders Nice find! Fixed that as well.

@michael-o Checked > 8, but there are no occurrences as far as I can see.

core-it-suite/src/test/resources/mng-7045/pom.xml
```diff
@@ -25,8 +25,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.8.1</version>
         <configuration>
-          <source>8</source>
-          <target>8</target>
+          <source>7</source>
```
mthmulders, line 28: Is that valid? I thought dropping the `1.` is allowed only from 8 onwards, but I'm not sure...

michael-o: Likely depends on the Java version used

MartinKanters: Good point, I was a bit too hasty there. For my JDK (OpenJDK) it worked with "7", but made it "1.7" just to be sure. It works with that as well.

PR #140, opened 2022-03-11
[MRESOLVER-247] IT update for resolver behaviour change

cstamas:
I tried to be least intrusive here (so I left old but added
new values w/ explanation).

Once MRESOLVER-247 is merged, and maven consumes it, this IT
will fail, as maven will resolve LESS then this IT expects.

This change adjusts numbers, and makes the two IT pass OK.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5669ReadPomsOnce.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5669ReadPomsOnce.java
@@ -45,6 +45,16 @@ public class MavenITmng5669ReadPomsOnce
         super( "[4.0.0-alpha-1,)" );
     }
 
+    /**
+     * "Original" resolver w/ DFS and w/o skipper: it did resolve things that were actually later unused.
+     */
+    private static final int EXPECTED_LOG_SIZE_DFS = 168;
+
+    /**
+     * "Modern" resolver w/ BDS and w/ skipper: it does not resolver unnecessary things.
+     */
+    private static final int EXPECTED_LOG_SIZE_BFS = 145;
+
     public void testWithoutBuildConsumer()
         throws Exception
     {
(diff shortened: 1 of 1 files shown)


michael-o (commented):
I don't understand this change. Why 140+? And how did `+ 4`  make it to `- 4`? Moving the expected value to a constant make sense, but the rest is unclear.

cstamas:
> I don't understand this change. Why 140+? And how did `+ 4` make it to `- 4`? Moving the expected value to a constant make sense, but the rest is unclear.

Changes one by one:
* The assert for 140+ is really just to exclude false positives (ie. Maven does not parse even one POM). We KNOW there are hundred-something POMs to parse, but we DO NOT WANT to fix that number (as will change between DFS/BFS/skipper)
* instead to assert lines in log (created by agent), we "remember" the count, then we make that list unique, and compare size again
* the +4 was factored out, the sum was asserted (168+4), so assertion happened with 172, hence, to get 168 out of 172 we do -4.
* In other words (applies to testWithBuildConsumer) method only: we KNOW that reactor POMs are read twice (and reactor has 2 POMs, hence we "correct" the count with -4)
* finally, I have a list of strings, making them unique by file path (using `model.building.source`) and asserting result is list.size - 1 = uniqList.size (as super pom is OMITTED from this list, it has NO `model.building.sourcre`).

Example of log entries for POM that went thru model builder vs Super POM (reformatted for easier read, they are all one liners):
```
{
  org.apache.maven.model.io.inputSource=null null, 
  org.apache.maven.model.io.isStrict=true, 
  org.apache.maven.model.building.source=/home/cstamas/Worx/apache-maven/maven-integration-testing/core-it- suite/target/test-classes/mng-5669-read-poms-once/module3/pom.xml
}, 
{
  org.apache.maven.model.io.inputSource=org.apache.maven:maven-model-builder:4.0.0-alpha-1-SNAPSHOT:super-pom jar:file:/home/cstamas/Worx/apache-maven/maven-integration-testing/core-it-suite/target/apache-maven/lib/maven-model-builder-4.0.0-alpha-1-SNAPSHOT.jar!/org/apache/maven/model/pom-4.0.0.xml, 
  xml:4.0.0=xml:4.0.0
}

```

michael-o:
Is this one still necessary?

PR #141, opened 2022-03-15
[MNG-7404] Drop support for prefixless expressions

mthmulders:
This PR drops support for expressions in the POM that do not start with `project.`.

Most existing IT's focussed solely on that, so their version range has been updated to end at 4.0.0-alpha-1. A few are mixed with other types of expressions, and they have been updated to remove the expressions that didn't start with `project.`.


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2339BadProjectInterpolationPre4Test.java
```diff
@@ -0,0 +1,75 @@
...
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-2339">MNG-2339</a>.
+ *
+ *
+ */
+public class MavenITmng2339BadProjectInterpolationPre4Test
+    extends AbstractMavenIntegrationTestCase
+{
+    public MavenITmng2339BadProjectInterpolationPre4Test()
+    {
+        super( "(2.0.8,4.0.0-alpha-1-SNAPSHOT)" ); // 2.0.9+
```
michael-o, line 36: The `SNAPSHOT` is not necessary. The magic will still work.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2339BadProjectInterpolationPre4Test.java
mthmulders: Of course!

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2339BadProjectInterpolationPre4Test.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.maven.it.util.ResourceExtractor;
+
+import java.io.File;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-2339">MNG-2339</a>.
+ *
+ *
+ */
+public class MavenITmng2339BadProjectInterpolationPre4Test
```
michael-o, line 31: This class isn't registered with the suite, how should this run?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng2339BadProjectInterpolationPre4Test.java
mthmulders: It would not. But adding it to the suite would, so I've pushed an additional commit just for that. Thanks for pointing that out. Double checked: running the IT's with Maven 3.8.5 runs both test classes for MNG-2339, running with Maven 4.0.0-alpha-1-SNAPSHOT runs only one and skips this newly added one.

michael-o: Have to considered to keep it in the original class and use this approach: https://github.com/apache/maven-integration-testing/blob/7fb56604e6e74afe293a42e82f20665566f10ba8/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4768NearestMatchConflictResolutionTest.java#L52?

mthmulders: Oh, cool, I didn't know that one! So no, certainly didn't consider that, but I'd say it's cleaner indeed (doesn't introduce a new class, etc.). I've restored the original situation with just that `requiresMavenVersion(...)` added. Thanks for the suggestion!

michael-o:
Will pick up tomrrow

michael-o (approved):

PR #143, opened 2022-04-01
[MNG-5222] IT test - warn about deprecated plugin params

slawekjaranowski:


slawekjaranowski:
https://github.com/apache/maven/pull/705

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
```diff
@@ -190,6 +190,7 @@ public static Test suite()
         suite.addTestSuite( MavenITmng5280SettingsProfilesRepositoriesOrderTest.class );
         suite.addTestSuite( MavenITmng5230MakeReactorWithExcludesTest.class );
         suite.addTestSuite( MavenITmng5224InjectedSettings.class );
+        suite.addTestSuite( MavenITmng5222MojoDeprectaeParamsTest.class );
```
kwin, line 193: ...MojoDeprecatedParamsTest.class

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
```diff
@@ -190,6 +190,7 @@ public static Test suite()
         suite.addTestSuite( MavenITmng5280SettingsProfilesRepositoriesOrderTest.class );
         suite.addTestSuite( MavenITmng5230MakeReactorWithExcludesTest.class );
         suite.addTestSuite( MavenITmng5224InjectedSettings.class );
+        suite.addTestSuite( MavenITmng5222MojoDeprecatedParamsTest.class );
```
michael-o, line 193: This should be as first since we have agreed that the ITs are in reverse chronological of fix date.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/IntegrationTestSuite.java
slawekjaranowski: ok, so sentence "it counts down to zero" it is not always true with - reverse chronological of fix date ... line 90 😄

michael-o: Correct, we had a discussion some time ago. We wanted recently fixed fail fast.

michael-o:
Call me stupid, but I ran the IT, but cannot see any warnings on the `log.txt`. I guess they aren't retained for further anlysis? This makes it hard to manually review output.

slawekjaranowski:
There are three test - one test `testEmptyConfiguration` confirm that no warning is generated when no configuration is used.

All test use the same test project and the same logs .... so when `testEmptyConfiguration` was  executed as last - log will not contains warnings.

michael-o:
Checking...

PR #151, opened 2022-04-16
[MNG-5222] Improvement in deprecated params detection

slawekjaranowski:
for https://github.com/apache/maven/pull/722

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5222MojoDeprecatedParamsTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5222MojoDeprecatedParamsTest.java
@@ -35,7 +35,7 @@ public class MavenITmng5222MojoDeprecatedParamsTest
 {
     public MavenITmng5222MojoDeprecatedParamsTest()
     {
-        super( "[3.9.0,)" );
+        super( "[4.0.0-alpha-1,)" );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/DeprecatedConfigMojo.java
```diff
@@ -47,6 +48,18 @@
...
+     * @deprecated bean read only
+     */
+    @Parameter(  defaultValue = "${project.artifact}", readonly = true )
+    private Artifact deprecatedBeanReadOnly;
+
+    /**
+     * @deprecated bean
+     */
+    @Parameter(  defaultValue = "${project.artifact}" )
+    private Artifact deprecatedBean;
```
michael-o, line 61: I don't see the corresponding test...

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5222MojoDeprecatedParamsTest.java
```diff
@@ -35,7 +35,7 @@
 {
     public MavenITmng5222MojoDeprecatedParamsTest()
     {
-        super( "[3.9.0,)" );
+        super( "[4.0.0-alpha-1,)" );
```
michael-o, line 38: Why back to 4 and not with 3.9?

slawekjaranowski: When this change will be merged with current fix to Maven master test on 3.9 will fail. 
I will restore it with together cherry pick/port change to 3.9.x branch of Maven core.

core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/DeprecatedConfigMojo.java
slawekjaranowski: It is corner case when internal value, project object is used as property without read only flags.
Such parameter should not generated warning.

michael-o: I see, there is one space too much

PR #154, opened 2022-04-28
[MNG-7464] Warn about using read-only parameters for Mojo in configuration

slawekjaranowski:


michael-o:
Question: So settings these is a no-op, true?
I wonder whether we'd like to have this warning in 3.9.0 as well.

pzygielo:
> Question: So settings these is a no-op, true?

How I checked it:
Set compiler's [`outputDirectory`](https://github.com/apache/maven-compiler-plugin/blob/1d912302d1f4a315b1db18fcc3197cae7796cc4f/src/main/java/org/apache/maven/plugin/compiler/CompilerMojo.java#L74). It is respected (at least in maven 3.8.5) and value from pom is used.

I'm 100% sure that _readonlines_ of parameter probably results in not having it in mojo docs only.

pzygielo:
MNG-5001

slawekjaranowski:
> I wonder whether we'd like to have this warning in 3.9.0 as well.

will be in next PR

slawekjaranowski:
is is first step for MAN-5001 ... as we see many plugins should be changed like m-compiler with `outputDirectory` 
if user need override some of parameters those should be not read-only, or plugins should provide another way ...
So only warring

pzygielo:
> is first step for MAN-5001

Indeed - and it's even linked to it. I was blind, but now I see.

michael-o (approved):

PR #159, opened 2022-05-12
Mute Jetty

cstamas:
The HttpServer helper is used by two ITs:
* MavenITmng7470ResolverTransportTest
* MavenITmng5771CoreExtensionsTest

And it logs in case of MNG7470 for unknown reason. My bet would
be some static initing happening as side effect of OTHER ITs that
use Jetty directly, like MavenITmng5868NoDuplicateAttachedArtifacts.

Anyway, this sorts out Jetty logging.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/HttpServer.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/HttpServer.java
@@ -11,6 +11,8 @@ import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.handler.AbstractHandler;
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.StdErrLog;
 import org.eclipse.jetty.util.security.Constraint;
 import org.eclipse.jetty.util.security.Password;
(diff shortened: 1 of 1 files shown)


cstamas:
@olamy any idea what is happening? I know Jetty 9.0 is ancient (2013), but still :smile:

olamy (approved):
jetty 9.0 ouchhh
see https://www.eclipse.org/jetty/download.php
and 9.4.x will be community EOL in June 2022
what about upgrading to at least 9.4?

slawekjaranowski:
> what about upgrading to at least 9.4?

many dependencies can update here ... 😄

cstamas:
> > what about upgrading to at least 9.4?
> 
> many dependencies can update here ... 😄

ITs are Java 7, Jetty 9.4 is Java 8, but many IT plugins and extensions are built with Maven 2.0 😭  So, this is a bit of a ratnest here

olamy:
> > > what about upgrading to at least 9.4?
> > 
> > 
> > many dependencies can update here ... 😄
> 
> ITs are Java 7, Jetty 9.4 is Java 8, but many IT plugins and extensions are built with Maven 2.0 😭 So, this is a bit of a ratnest here

Maven 2.x is an Antique EOL https://maven.apache.org/maven-2.x-eol.html (which says 2009 :) )
that's a bit of job I reckon but seriously we could say 1.8...

PR #161, opened 2022-05-16
Fix the MavenITmng7470ResolverTransportTest

cstamas:
Seems the test must be forked to work in, as it fails to work
in embedded mode.


michael-o:
What exactly will this fix?

cstamas:
There is some classloading issue that bogs logging (among other things, @slawekjaranowski  is looking). By forking invocation in these tests, they become isolated, hence, will not suffer from these and DEBUG logging of resolver is back. Currently this IT fails as resolver does not emit DEBUG logs.
This is a "crude" fix, but goal is to make IT work/pass, not fail every commit in every 3.9+  maven.

slawekjaranowski (approved):

PR #162, opened 2022-05-17
[MNG-7487] Fix deadlock during forked lifecycle executions

gnodet:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7433DeadlockTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.maven.it.util.ResourceExtractor;
+
+public class MavenITmng7433DeadlockTest extends AbstractMavenIntegrationTestCase
+{
+    private static final String PROJECT_PATH = "/mng-7433-deadlock";
+
+    public MavenITmng7433DeadlockTest()
+    {
+        super( "[3.8.4,)" );
```
michael-o, line 34: Does this verify that there is a deadlock or your locks avoid it? Because the version says 3.8.4.

gnodet: The test will just hang forever if running with 3.8.5.  It can be run with any other version afaik.

core-it-suite/src/test/resources/mng-7433-deadlock/plugin/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+      <artifactId>maven-plugin-annotations</artifactId>
+      <version>${maven-version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-plugin-api</artifactId>
+      <version>${maven-version}</version>
+      <scope>provided</scope>
+    </dependency>
```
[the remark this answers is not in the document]
gnodet: Fixed

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7433DeadlockTest.java
michael-o: Then the range must be fixed for: `(,3.8.4],[3.8.6,)`

gnodet: Fixed

michael-o:
Moved message helper to https://github.com/apache/maven/pull/746.

michael-o (approved):
Nice work!

PR #164, opened 2022-05-19
Upgrade maven-verifier to 1.8.0

slawekjaranowski:
- upgrade maven-verifier
- remove code implemented in maven-verifier
- replace deprecated methods


olamy:
just figured out we do not have any default checks here?
maybe some simple runs with last maven core release?

slawekjaranowski:
When we create branch with the same name in Maven core build will be triggered 😄
 
https://github.com/apache/maven/actions?query=branch%3Amaven-verifier-1.8.0

cstamas (approved):

PR #165, opened 2022-05-21
[MNG-7468] Check unsupported plugins parameters in configuration

slawekjaranowski:


core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/TouchMojo.java
```diff
@@ -0,0 +1,82 @@
...
+            {
+                dir.mkdirs();
+            }
+
+            File touch = new File( dir, file );
+
+            try ( OutputStreamWriter w = new OutputStreamWriter(
+                new FileOutputStream( touch, true ), StandardCharsets.UTF_8 ) )
+            {
+                w.write( file );
```
michael-o, line 73: A `touch` produces a zero byte file, this one don't.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7468UnsupportedPluginsParametersTest.java
```diff
@@ -0,0 +1,183 @@
...
+        Verifier verifier = newVerifier( new File( testDir, project ).getAbsolutePath() );
+        verifier.executeGoal( "validate" );
+        verifier.verifyErrorFreeLog();
+        verifier.resetStreams();
+
+        List<String> logLines = verifier.loadFile( verifier.getBasedir(), verifier.getLogFileName(), false );
+        return findUnknownWarning( logLines );
+    }
+
+    private void assertWarningPresents( List<String> warnLines )
```
michael-o, line 158: Rather `assertWarningContains`, no?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7468UnsupportedPluginsParametersTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+        assertTrue( warnLines.remove(
+            "[WARNING] Parameter 'invalidXml' is unknown for plugin 'maven-it-plugin-configuration:2.1-SNAPSHOT:touch (default)'" ) );
+
+        assertTrue( "Not verified line: " + warnLines, warnLines.isEmpty() );
+    }
+
+    private List<String> findUnknownWarning( List<String> logLines )
+    {
+        Pattern pattern = Pattern.compile( "\\[WARNING] Parameter .* is unknown.*" );
```
michael-o, line 171: WE rather require content: `.+`, no?

core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/TouchMojo.java
```diff
@@ -0,0 +1,73 @@
...
+    {
+        try
+        {
+            if ( !dir.exists() )
+            {
+                dir.mkdirs();
+            }
+
+            File touch = new File( dir, file );
+            touch.createNewFile();
```
michael-o, line 66: https://www.youtube.com/watch?v=_NNYI8VbFyY

core-it-suite/src/test/resources/mng-7468-unsupported-params/config-build-execution/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+          <execution>
+            <goals>
+              <goal>touch</goal>
+            </goals>
+            <configuration>
+              <invalidValue>value</invalidValue>
+              <invalidXml>
+                <item>value</item>
+              </invalidXml>
+            </configuration>
```
michael-o, line 50: This confuses me a bit. We don't validate the value, but only the param, no? So why not `invalidParam`? What am I missing?
The next block is an Xpp3Dom, so I guess `invalidXml` is correct.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7468UnsupportedPluginsParametersTest.java
slawekjaranowski: in this case will be the same ... we searching lines of this pattern and in next step specific line values are compared

michael-o: OK, I see.

core-it-suite/src/test/resources/mng-7468-unsupported-params/config-build-execution/pom.xml
slawekjaranowski: ok, `invalidParam` will be more precise - yes we check parameter names not values

`invalidXml` can be configuration for complex custom object or collections like list

I don't now  - can plugin have access to raw `Xpp3Dom` if not defined parameter in plugin.xml descriptor

core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/TouchMojo.java
michael-o: No, just wanted to add some fun. :-D

core-it-suite/src/test/resources/mng-7468-unsupported-params/config-build-execution/pom.xml
slawekjaranowski: both new?
 - `invalidValue` -> `invalidParam`
 - ?

slawekjaranowski: done

PR #167, opened 2022-05-23
[MNG-7360] Add an IT for build consumer with tags in plugin configs

gnodet:


core-it-suite/src/test/resources/mng-7360-build-consumer/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.apache.maven.its.mng7360</groupId>
+    <artifactId>parent</artifactId>
+    <version>1.0</version>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <jdk.version>11</jdk.version>
```
michael-o, line 30: Is this Java 11 specific? If yes, this test will fail everywhere Java 8 is used.

gnodet: The only goal run is the clean one, so I doubt this matters at all.  We can get rid of this line if you prefer.

michael-o: If it does not matter, then remove please.

gnodet: The pom is much slimmer now.

michael-o:
Indeed, pure minimalism.

gnodet:
> Indeed, pure minimalism.

A bit too much.  The plugin needs to be resolved for any goal to succeed... I had to use a real one.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7360BuildConsumer.java
```diff
@@ -0,0 +1,51 @@
...
+    public MavenITmng7360BuildConsumer()
+    {
+        super( "[4.0.0-alpha-1,)" );
+    }
+
+    public void testSelectModuleByCoordinate() throws Exception
+    {
+        final File projectDir = ResourceExtractor.simpleExtractResources( getClass(), PROJECT_PATH );
+        final Verifier verifier = newVerifier( projectDir.getAbsolutePath() );
+        verifier.executeGoals( Arrays.asList( "clean" ) );
```
mthmulders, line 47: Nit: I think even "validate" would have been enough, but this won't hurt either.

mthmulders (approved):

PR #168, opened 2022-05-24
[MNG-7443] IT for consistent logging of optional selectors

Giovds:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7443ConsistencyOfOptionalProjectsAndProfilesTest.java
```diff
@@ -0,0 +1,69 @@
...
+        verifier.executeGoals( Arrays.asList( "clean", "verify" ) );
+
+        final List<String> logLines = verifier.loadFile( verifier.getBasedir(), verifier.getLogFileName(), false );
+
+        int projectSelectorMissingCounter = 0;
+        int profileSelectorMissingCounter = 0;
+
+        for ( String logLine : logLines )
+        {
+            if ( logLine.equals( "[INFO] The requested optional projects [:does-not-exist] do not exist." ) )
```
mthmulders, line 55: I would try to match on the most relevant part of the log line, e.g. "requested optional project" and ":does not exist". The other parts are not so important. Same goes for line 60.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7443ConsistencyOfOptionalProjectsAndProfilesTest.java
```diff
@@ -0,0 +1,72 @@
...
+        super( "[4.0.0-alpha-1,)" );
+    }
+
+    public void testConsistentLoggingOfOptionalProfilesAndProjects() throws IOException, VerificationException
+    {
+        File testDir = ResourceExtractor.simpleExtractResources( getClass(),
+                "/mng-7443-consistency-of-optional-profiles-and-projects" );
+
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+        verifier.addCliOption( "-pl ?:does-not-exist" );
```
michael-o, line 43: These are two options, hence the space.

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7443ConsistencyOfOptionalProjectsAndProfilesTest.java
mthmulders: Maybe the same for the line below this one, which now reads `-P?does-not-exist-either`...

michael-o:
Running this now, the IT needs some non-functional improvements I will perform

PR #174, opened 2022-06-19
[MNG-7504] Don't print warning unsupported reportPlugins for m-site-p

slawekjaranowski:


slawekjaranowski:
After MNG-7505 can be activated also for 4.x

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7504NotWarnUnsupportedReportPluginsTest.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author Slawomir Jaranowski
+ */
+public class MavenITmng7504NotWarnUnsupportedReportPluginsTest extends AbstractMavenIntegrationTestCase
+{
+    private static final String PROJECT_PATH = "/mng-7504-warn-unsupported-report-plugins";
+
+    public MavenITmng7504NotWarnUnsupportedReportPluginsTest()
+    {
+        super( "[3.9.0]" );
```
michael-o, line 41: I think you can even drop the brackets.

michael-o: Stupid question, why only 3.9.0? My understanding:
* 3.9.0 is suppressing this
* 4.0.0 does not even invoke

In both cases the warning should be visible, no?

slawekjaranowski: currently only 3.9.0 has a fix - https://github.com/apache/maven/pull/759
4.0.0 - still print warning MNG-7505

michael-o (approved):

PR #175, opened 2022-06-21
Relax test for wider terminals

gnodet:
Maven 4.0.x has some code new code which takes care of the terminal width to adjust the output.  The number of dots can thus vary when running the tests locally.


michael-o:
Do I need to make my terminal full screen to make it work?

gnodet:
> Do I need to make my terminal full screen to make it work?

No, I think any width > 80 will fail the test for maven 4.0.x branch if you run them from a real terminal.

michael-o:
> > Do I need to make my terminal full screen to make it work?
> 
> No, I think any width > 80 will fail the test for maven 4.0.x branch if you run them from a real terminal.

Interesting.. I have been running PuTTY fullscreen for a very long time, no failures:
```
osipovmi@deblndw011x:~/var/Projekte/maven-integration-testing (bootstrap *=)
$ tput cols
135
osipovmi@deblndw011x:~/var/Projekte/maven-integration-testing (bootstrap *=)
$ tput lines
30
```

PR #177, opened 2022-06-27
[MNG-7474] SessionScoped beans should be singletons for a given session

gnodet:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7474SessionScopeTest.java
```diff
@@ -0,0 +1,64 @@
...
+        throws Exception
+    {
+        File testDir = ResourceExtractor.simpleExtractResources( getClass(), "/mng-7474-session-scope" );
+        projectDir = new File( testDir, "project" );
+    }
+
+    public void testSessionScope()
+            throws Exception
+    {
+        File testDir = ResourceExtractor.simpleExtractResources( getClass(), "/mng-7474-session-scope" );
```
michael-o, line 54: Why another extract?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7474SessionScopeTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        Verifier verifier = newVerifier( testDir.getAbsolutePath() );
+        verifier.executeGoal( "validate" );
+        verifier.verifyErrorFreeLog();
+        verifier.resetStreams();
+
+        List<String> logLines = verifier.loadFile( verifier.getBasedir(), verifier.getLogFileName(), false );
```
michael-o, line 61: Why do you load them w/o any verification?

michael-o:
Running ITs...

michael-o:
After version bump of MRESOURCES all tests pass for me on Windows 10 and FreeBSD 12. Waiting for comment answers.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7474SessionScopeTest.java
gnodet: It was for debugging iirc. I've removed it.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7474SessionScopeTest.java
gnodet: Right, unneeded.

PR #186, opened 2022-07-19
Fix minor typos

marcwrobel:
Fix various minor typos in :

- project documentations,
- javadoc and comments,
- test class and method names,
- exception messages.


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3023ReactorDependencyResolutionTest.java
```diff
@@ -80,7 +80,7 @@ public void testitMNG3023A() throws Exception
     /**
      * Test that reactor projects are included in dependency resolution.
      *
-     * I this pass, the dependency artifact should have the file $(basedir)/dependency/dependency-classes
+     * If this pass, the dependency artifact should have the file $(basedir)/dependency/dependency-classes
```
michael-o, line 83: passes?

core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/ConfigMojo.java
```diff
@@ -118,7 +118,7 @@
     /**
      * A simple parameter of type {@link java.lang.Integer}.
      *
-     * @parameter property="config.intergerParam"
+     * @parameter property="config.integerParam"
```
michael-o, line 121: Attention: Is this property used in the IT?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3023ReactorDependencyResolutionTest.java
marcwrobel: Of course ! Fixed.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3023ReactorDependencyResolutionTest.java
```diff
@@ -112,7 +112,7 @@ public void testitMNG3023B()
     /**
      * Test that reactor projects are included in dependency resolution.
      *
-     * I this pass, the dependency should have been installed, so the dependency artifact should have
+     * If this pass, the dependency should have been installed, so the dependency artifact should have
```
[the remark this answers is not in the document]
marcwrobel: Fixed.

core-it-support/core-it-plugins/maven-it-plugin-configuration/src/main/java/org/apache/maven/plugin/coreit/ConfigMojo.java
marcwrobel: It did not find any reference in this repo. And according to https://github.com/search?q=config.intergerParam&type=code there is no references in other repos either.

Do you think it's safer to roll back the change ?

michael-o: Use grep, then we are sure.

marcwrobel: No result when grepping on `intergerParam`.

PR #190, opened 2022-08-22
[MNG-7353] Add missing dependencies to bootstrap.txt

hgschmie:
Add missing dependencies to bootstrap.txt instead of changing the test internal versions (reverts 8dd42243ecf74c26d4a2788e7871e660eef77115).

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7353CliGoalInvocationTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7353CliGoalInvocationTest.java
@@ -49,13 +49,13 @@ public class MavenITmng7353CliGoalInvocationTest
     public void testPrefixGoal()
         throws Exception
     {
-        run( "pluginPrefix-goal", "dependency:list", "maven-dependency-plugin:2.7:list (default-cli)" );
+        run( "pluginPrefix-goal", "dependency:list", "maven-dependency-plugin:3.3.0:list (default-cli)" );
     }
 
     public void testPrefixGoalAtId()
         throws Exception
     {
-        run( "pluginPrefix-goal@id", "dependency:list@id", "maven-dependency-plugin:2.7:list (id)" );
+        run( "pluginPrefix-goal@id", "dependency:list@id", "maven-dependency-plugin:3.3.0:list (id)" );
     }
 
     /**
@@ -64,7 +64,7 @@ public class MavenITmng7353CliGoalInvocationTest
     public void testPrefixVersionGoal()
         throws Exception
     {
-        run( "pluginPrefix-version-goal", "dependency:2.8:list", "maven-dependency-plugin:2.8:list (default-cli)" );
+        run( "pluginPrefix-version-goal", "dependency:3.1.1:list", "maven-dependency-plugin:3.1.1:list (default-cli)" );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


hgschmie:
@hboutemy @slawekjaranowski This is an alternate PR to fix the 7353 problem by adding the missing deps to the bootstrap.

slawekjaranowski (changes requested):
I would consider to one of mock plugins from:
https://github.com/apache/maven-integration-testing/tree/master/core-it-support/core-it-plugins
or create next mock plugin

After we release next version of maven-dependency-plugin the situation will be the same.

hgschmie:
The currently breaking tests refer to those two versions explicitly. So bootstrapping them into the local repo should do the trick. Not sure how a new release will break this. I leave the decision to make changes to the actual tests to @hboutemy.

slawekjaranowski:
IMHO `dependency:list` will be resolved to the latest available version, which will not present in bootstrap

hgschmie:
My understanding is that the failing test is trying to run the plugin with explicit versions (that is what is in the test code) and those are 3.1.1 and 3.3.0

I did the trick that @mthmulders explained for the integration test (create a branch with the same name as the PR branch in maven) and with this patch applied, it passes the integration tests: https://github.com/apache/maven/runs/7958258759 (well the MacOS tests failed but for a different reason), but the Linux and windows pass and they never did before. So that seems to work.

hgschmie:
https://github.com/apache/maven-integration-testing/blob/2181675286bf085c49198858d6bd0f65178d17b2/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7353CliGoalInvocationTest.java contains explicit versions for goal invocation (I think that is the point of the test). The execution *may* yield different versions in its output but I don't think that this is relevant.

slawekjaranowski (approved):
ok, we can merge 
my other build suffer from missing m-clean-p version

michael-o:
I am trying to reproduce the problem here...

michael-o:
@hgschmie Can you share the error output you see? I cannot reproduce the failure. It works for me with an empty repo.

slawekjaranowski:
@michael-o look this: https://github.com/apache/maven/actions/workflows/maven.yml
more of build failed of missing m-clean-p

slawekjaranowski:
locally with Maven from master I have the same

```
[INFO] --< org.apache.maven.its.mng7443:consistent-logging-of-optional-projects-and-profiles-selectors >--
[INFO] Building consistent-logging-of-optional-projects-and-profiles-selectors 1.0
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] Downloading from central: file:target/null/org/apache/maven/plugins/maven-clean-plugin/3.1.0/maven-clean-plugin-3.1.0.pom
[WARNING] The POM for org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 is missing, no dependency information available
[INFO] Downloading from central: file:target/null/org/apache/maven/plugins/maven-clean-plugin/3.1.0/maven-clean-plugin-3.1.0.jar
[INFO] ------------------------------------------------------------------------
...
[ERROR] Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Could not find artifact org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 in central (file:target/null) -> [Help 1]
org.apache.maven.plugin.PluginResolutionException: Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Could not find artifact org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 in central (file:target/null)

```

michael-o:
> locally with Maven from master I have the same
> 
> ```
> [INFO] --< org.apache.maven.its.mng7443:consistent-logging-of-optional-projects-and-profiles-selectors >--
> [INFO] Building consistent-logging-of-optional-projects-and-profiles-selectors 1.0
> [INFO]   from pom.xml
> [INFO] --------------------------------[ jar ]---------------------------------
> [INFO] Downloading from central: file:target/null/org/apache/maven/plugins/maven-clean-plugin/3.1.0/maven-clean-plugin-3.1.0.pom
> [WARNING] The POM for org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 is missing, no dependency information available
> [INFO] Downloading from central: file:target/null/org/apache/maven/plugins/maven-clean-plugin/3.1.0/maven-clean-plugin-3.1.0.jar
> [INFO] ------------------------------------------------------------------------
> ...
> [ERROR] Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Could not find artifact org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 in central (file:target/null) -> [Help 1]
> org.apache.maven.plugin.PluginResolutionException: Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Could not find artifact org.apache.maven.plugins:maven-clean-plugin:jar:3.1.0 in central (file:target/null)
> ```

Weird why it does not fail for me with two setups. I agree for the clean plugin, but what's with dependency plugin?

slawekjaranowski:
m-dependency-p - versions from test was added to bootstrap

michael-o (approved):

hboutemy:
The ITs don't at all use the fact that any version is LATEST, then using 2.7 and 2.8 instead of 3.0.0 and 3.1.1 is strictly the same result (I know, I wrote initial IT)

in general, adapting ITs to already available content is just a good practice: I just forgot to disable my local proxy before testing locally, then did not see it early
of course, downloading more content also work: but why download more when existing content can be sufficient?

I know my comment comes too late, there is no need to revert, because there is no strong consequence

PR #192, opened 2022-09-02
[MNG-7474] Add and fix IT

gnodet:
The IT was added but not referenced from the test suite, meaning it is never run.
This PR adds the test to the suite and actually fixes it because it was failing.


slawekjaranowski (approved):

michael-o:
I will review this.

michael-o (approved):
Made the last open change. Good to merge.

PR #176, opened 2022-06-23
Move ITs to JDK 8 / JUnit 5

gnodet:
This requires https://github.com/apache/maven-verifier/pull/31

PR #197, opened 2022-09-09
Remove verifier.getCliOptions().add( x ) pattern

gnodet:
Follow up to ecf4a04cd1c4b7698f28e3332cd24c23116c1467 which missed a few tests using the above pattern

### DIFF
--- a/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
+++ b/core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
@@ -571,14 +571,14 @@ public abstract class AbstractMavenIntegrationTestCase
             // dedicated CLI option only available since MNG-3914
             if ( matchesVersionRange( "[2.1.0,)" ) )
             {
-                verifier.getCliOptions().add( "--global-settings" );
+                verifier.addCliOption( "--global-settings" );
                 if ( path.indexOf( ' ' ) < 0 )
                 {
-                    verifier.getCliOptions().add( path );
+                    verifier.addCliOption( path );
                 }
                 else
                 {
-                    verifier.getCliOptions().add( '"' + path + '"' );
+                    verifier.addCliOption( '"' + path + '"' );
                 }
             }
             else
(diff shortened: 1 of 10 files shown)


michael-o:
Good catch. @slawekjaranowski This list should be read only.

core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
```diff
@@ -592,7 +592,7 @@ protected Verifier newVerifier( String basedir, String settings, boolean debug )
             // Java7 TLS protocol
             if ( VersionRange.createFromVersionSpec( "(,1.8.0)" ).containsVersion( getJavaVersion() ) )
```
michael-o, line 593: This is a good catch. Stuff like this can be removed from master because we have Java 8+

michael-o (approved):

slawekjaranowski:
> Good catch. @slawekjaranowski This list should be read only.

exactly

gnodet:
> > Good catch. @slawekjaranowski This list should be read only.
> 
> exactly

Note that there's one test which uses `verifier.getCliOptions().remove( ... )` !

core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
gnodet: Not yet afaik. This will be the case once https://github.com/apache/maven-integration-testing/pull/176 is merged.

slawekjaranowski (approved):

michael-o:
> > > Good catch. @slawekjaranowski This list should be read only.
> > 
> > 
> > exactly
> 
> Note that there's one test which uses `verifier.getCliOptions().remove( ... )` !

WHAT? That needs to be analyzed...

gnodet:
> > > > Good catch. @slawekjaranowski This list should be read only.
> > > 
> > > 
> > > exactly
> > 
> > 
> > Note that there's one test which uses `verifier.getCliOptions().remove( ... )` !
> 
> WHAT? That needs to be analyzed...

https://github.com/apache/maven-integration-testing/blob/4aec5c3df75987ca4c8683b02f5b90d7462b07c2/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4189UniqueVersionSnapshotTest.java#L79-L80

slawekjaranowski:
New instance of Verifier should be used for each executions - Verifier should be immutable

michael-o:
> New instance of Verifier should be used for each executions - Verifier should be immutable

Yes, this makes it idiotproof.

PR #199, opened 2022-09-14
Remove code corresponding to Maven 2.x and JDK 7

slawekjaranowski:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0294MergeGlobalAndUserSettingsTest.java
```diff
@@ -53,16 +53,9 @@ public void testitMNG294()
...
-        else
-        {
-            verifier.setSystemProperty( "org.apache.maven.global-settings", "global-settings.xml" );
-        }
+        verifier.addCliOption( "--global-settings" );
+        verifier.addCliOption( "global-settings.xml" );
+
         if ( matchesVersionRange( "(3.2.5,)" ) )
         {
             verifier.getSystemProperties().put( "maven.multiModuleProjectDirectory", testDir.getAbsolutePath() );
```
michael-o, line 61: Can't this be simplified as well since we have now 3.9.0+?

core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
```diff
@@ -568,33 +568,19 @@ protected Verifier newVerifier( String basedir, String settings, boolean debug )
...
-                {
-                    verifier.addCliOption( '"' + path + '"' );
-                }
+                verifier.addCliOption( path );
             }
             else
             {
-                verifier.getSystemProperties().put( "org.apache.maven.global-settings", path );
+                verifier.addCliOption( '"' + path + '"' );
             }
```
michael-o, line 579: Do we still need to quote ourselves? Feels weird, no?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng0294MergeGlobalAndUserSettingsTest.java
slawekjaranowski: Even more `multiModuleProjectDirectory` is set by Verifier 1.8.0

core-it-support/maven-it-helper/src/main/java/org/apache/maven/it/AbstractMavenIntegrationTestCase.java
slawekjaranowski: I wouldn't like touch it until Verifier 2.0

gnodet: Yes, note that the upgrade to verifier 2.0 will break quotes arguments.  Also note that I plan to merge https://github.com/apache/maven-integration-testing/pull/176 soon and it includes an upgrade to verifier 2.0-SNAPSHOT.  Unless we really do want to postpone until it's released, but the snapshots are available on public apache repos, so that's not a dealbreaker, especially as ITs are never released.

slawekjaranowski: We have many changes in Verifier 2.0.0, many deprecated methods, etc 
My proposition is to release Verifier 2.0.0-M1 as is now - I can do it in next weekend.

In next step after #176   we can clean of using deprecated methods we can release next milestone with next part of resolved issues and so on, until to final 2.0.0

As always I like step by step and if it possible split bigger changes.

What do you think?

gnodet: Sounds like a plan to me !

PR #209, opened 2022-11-11
[MNG-7338] Disable CI mode for certain tests which verify on log statements.

MartinKanters:
Relevant [Maven PR](https://github.com/apache/maven/pull/869)

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4461ArtifactUploadMonitorTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4461ArtifactUploadMonitorTest.java
@@ -47,6 +47,7 @@ public class MavenITmng4461ArtifactUploadMonitorTest extends AbstractMavenIntegr
         Verifier verifier = newVerifier(testDir.getAbsolutePath());
         verifier.setAutoclean(false);
         verifier.deleteDirectory("target");
+        verifier.setEnvironmentVariable("CI", "false");
         verifier.addCliArgument("validate");
         verifier.execute();
         verifier.verifyErrorFreeLog();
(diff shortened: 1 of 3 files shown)


michael-o:
I don't really think it is not in the spirit of this var. As far as I understand it is either `CI=true` or not present at all.

MartinKanters:
> I don't really think it is not in the spirit of this var. As far as I understand it is either `CI=true` or not present at all.

I agree. I tried to remove it using `verifier.getEnvironmentVariables().remove( "CI" );`, but that doesn't work unfortunately. 
I would not like to change the actual environment variables on the machine if possible, so I thought this was the best way forward.

rmannibucau:
@michael-o it is the defined semantic now but was not originally and it was soletimes used to switch between vendors. That said can not be critical due to the impact.

slawekjaranowski:
> I agree. I tried to remove it using `verifier.getEnvironmentVariables().remove( "CI" );`, but that doesn't work unfortunately.

Should work with:

```
verifier.addEnvironemnt( "CI", null );
```

but needs a maven-sahred-utils 3.3.4 ... as I see 0.9 is used in tests 😄

MartinKanters:
> > I agree. I tried to remove it using `verifier.getEnvironmentVariables().remove( "CI" );`, but that doesn't work unfortunately.
> 
> Should work with:
> 
> ```
> verifier.addEnvironemnt( "CI", null );
> ```
> 
> but needs a maven-sahred-utils 3.3.4 ... as I see 0.9 is used in tests 😄

Right.. not sure if that's worth the effort for this change haha. I tried it quickly but didn't see an addEnvironment variable method. I did see that invoking setEnvironment with a `null` value should remove it from the environment variables. But that didn't work before. I think that's because those environment vars are probably merged later with the actual environment variables (which would add CI=true again)

PR #210, opened 2022-11-11
Allow basedir system property to be null for build info.txt

slawekjaranowski:
- basedir can be null when test is executed by IDE, use default value
- use try-with-resources

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
@@ -27,7 +27,6 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.apache.maven.shared.verifier.Verifier;
-import org.codehaus.plexus.util.IOUtil;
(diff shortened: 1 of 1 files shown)


core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
```diff
@@ -42,46 +44,34 @@ public class TestSuiteOrdering implements ClassOrderer
 
     final Map<Class<?>, Integer> tests = new HashMap<>();
 
-    private static void infoProperty( PrintStream info, String property )
+    private static void infoProperty( Writer info, String property ) throws IOException
     {
-        info.println( property + ": " + System.getProperty( property ) );
+        info.append( property ).append( ": " ).append( System.getProperty( property ) );
```
michael-o, line 49: Wthere is the line separator?

slawekjaranowski: revert to PrintStrem ... not need improvement

michael-o (approved):

PR #211, opened 2022-11-11
Fix flappy test - mng-6720-fail-fast

slawekjaranowski:
- old m-resource-p can fail in multi thread executions
- test project use aggregation - so we need provided configuration in each module
- replace deprecated method
- add message for assertions

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6720FailFastTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6720FailFastTest.java
@@ -19,14 +19,12 @@ package org.apache.maven.it;
  * under the License.
  */
 
-import org.apache.maven.shared.verifier.util.ResourceExtractor;
-import org.apache.maven.shared.verifier.Verifier;
-import org.apache.maven.shared.verifier.VerificationException;
-
 import java.io.File;
-import java.util.Arrays;
 import java.util.List;
 
+import org.apache.maven.shared.verifier.VerificationException;
+import org.apache.maven.shared.verifier.Verifier;
+import org.apache.maven.shared.verifier.util.ResourceExtractor;
 import org.junit.jupiter.api.Test;
 
 /**
@@ -36,7 +34,7 @@ import org.junit.jupiter.api.Test;
  * <a href="https://issues.apache.org/jira/browse/MNG-6720">MNG-6720</a>.
  *
  */
-public class MavenITmng6720FailFastTest
+class MavenITmng6720FailFastTest
     extends AbstractMavenIntegrationTestCase
 {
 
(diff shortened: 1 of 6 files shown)


slawekjaranowski:
example stack trace:

```
Caused by: org.apache.maven.plugin.PluginExecutionException: 
   Execution default-resources of goal org.apache.maven.plugins:maven-resources-plugin:2.6:resources failed.
...
Caused by: java.util.ConcurrentModificationException
    at java.util.Hashtable$Enumerator.next (Hashtable.java:1387)
    at java.util.Hashtable.putAll (Hashtable.java:523)
    at org.apache.maven.shared.filtering.DefaultMavenFileFilter.getDefaultFilterWrappers (DefaultMavenFileFilter.java:198)
    at org.apache.maven.shared.filtering.DefaultMavenResourcesFiltering.filterResources (DefaultMavenResourcesFiltering.java:161)
    at org.apache.maven.plugin.resources.ResourcesMojo.execute (ResourcesMojo.java:301)

```

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6720FailFastTest.java
```diff
@@ -36,7 +34,7 @@
  * <a href="https://issues.apache.org/jira/browse/MNG-6720">MNG-6720</a>.
  *
  */
-public class MavenITmng6720FailFastTest
+class MavenITmng6720FailFastTest
```
michael-o, line 37: Is the rest of our tests also package private?

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6720FailFastTest.java
```diff
@@ -46,35 +44,34 @@ public MavenITmng6720FailFastTest()
...
 
         Verifier verifier = newVerifier( testDir.getAbsolutePath(), false );
-        verifier.setMavenDebug( false );
         verifier.setAutoclean( false );
-        verifier.addCliOption( "-T" );
-        verifier.addCliOption( "2" );
-        verifier.addCliOption( "-Dmaven.test.redirectTestOutputToFile=true" );
+        verifier.addCliArguments( "-T", "2" );
+        verifier.addCliArgument( "-Dmaven.test.redirectTestOutputToFile=true" );
+        verifier.addCliArguments( "clean", "test" );
```
michael-o, line 57: Beautiful, this new API :-D

core-it-suite/src/test/resources/mng-6720-fail-fast/pom.xml
```diff
@@ -35,15 +35,4 @@ under the License.
...
-    <pluginManagement>
-      <plugins>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-resources-plugin</artifactId>
-          <version>3.2.0</version>
-        </plugin>
-      </plugins>
-    </pluginManagement>
-  </build>
```
michael-o, line 48: Why not use plugin management in the part for those?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng6720FailFastTest.java
slawekjaranowski: No yet, it is Junit5 convention ,so step by step

core-it-suite/src/test/resources/mng-6720-fail-fast/pom.xml
slawekjaranowski: because project use aggregation and child poms don't have defined parent - it is main reason for change

michael-o: Ah, yes. Make sense.

michael-o (approved):

PR #216, opened 2022-12-02
MNG-7608] Make native transport the default

cstamas:
Make ITs not blindly expect Wagon as transport, but be aware of change.

---

https://issues.apache.org/jira/browse/MNG-7608


gnodet (approved):

cstamas:
This PR may not be needed at all..... DO NOT MERGE

gnodet:
> This PR may not be needed at all..... DO NOT MERGE

Btw, it would be nice to split the verifier upgrade from the actual changes.
And I don't see how this would not be necessary given the name of the transporter is checked in the debug log output...

cstamas:
True 7470 is needed, but the other ones are not, gimme some....

PR #221, opened 2023-01-01
Upgrade Groovy in one of the ITs to support JDK 19

psiroky:
This should be the only change needed to run the tests on JDK 19. Tested with Maven 3.8.7 and latest master as of today (4.0.0-alpha-4-SNAPSHOT):
 * `mvn clean install -Prun-its -Dmaven.repo.local=$(pwd)/repo `
 * `mvn clean install -Prun-its -Dmaven.repo.local=$(pwd)/repo -Dmaven.home=<path-to-maven-git-repo>/apache-maven/target/apache-maven-4.0.0-alpha-4-SNAPSHOT`

### DIFF
--- a/core-it-suite/src/test/resources-filtered/bootstrap.txt
+++ b/core-it-suite/src/test/resources-filtered/bootstrap.txt
@@ -12,8 +12,8 @@ javax.annotation:javax.annotation-api:1.2
 junit:junit:4.12
 org.apache.ant:ant:1.10.8
 org.apache.geronimo.specs:geronimo-jcdi_2.0_spec:1.3
-org.apache.groovy:groovy-ant:4.0.0-alpha-3
-org.apache.groovy:groovy:4.0.0-alpha-3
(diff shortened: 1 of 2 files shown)


psiroky:
@slawekjaranowski thanks for looking at the PR / approving the workflow (at least I hope it was you). I rebased on top of latest master, which contains your fix to run the build with Maven 3.8.7. That should hopefully make the it pass. Please approve the workflow once more when you get a chance.

psiroky:
Just thinking out loud: would it make sense to add a Github build running with latest released JDK? So 19 at this moment, and 20 once it gets released (and once the build works there). What do you think?

slawekjaranowski:
> Just thinking out loud: would it make sense to add a Github build running with latest released JDK? So 19 at this moment, and 20 once it gets released (and once the build works there). What do you think?

We should use similar JDKs for this project and for Maven itself ... 
Also it is open question which JKD use for plugins build ...

But we can try here with JDK 19 ... please check locally first

PR #223, opened 2023-01-03
[MNG-7661] Use provided scope for Maven deps in IT plugins

psiroky:
Fixes these kinds of warnings:
```
[INFO] --- maven-plugin-plugin:3.6.4:descriptor (default-descriptor) @ maven-it-plugin-bootstrap ---
[ERROR] Some dependencies of Maven Plugins are expected to be in provided scope.
Please make sure that dependencies listed below declared in POM
have set '<scope>provided</scope>' as well.The following dependencies are in wrong scope:
 * org.apache.maven:maven-plugin-api:jar:3.8.6:compile
 * org.apache.maven:maven-model:jar:3.8.6:compile
 * org.apache.maven:maven-artifact:jar:3.8.6:compile
 * org.apache.maven:maven-core:jar:3.8.6:compile
 * org.apache.maven:maven-settings:jar:3.8.6:compile
 * org.apache.maven:maven-settings-builder:jar:3.8.6:compile
 * org.apache.maven:maven-builder-support:jar:3.8.6:compile
 * org.apache.maven:maven-repository-metadata:jar:3.8.6:compile
 * org.apache.maven:maven-model-builder:jar:3.8.6:compile
 * org.apache.maven:maven-resolver-provider:jar:3.8.6:compile
```

### DIFF
--- a/core-it-support/core-it-plugins/maven-it-plugin-settings/src/main/java/org/apache/maven/plugin/coreit/SettingsReadItMojo.java
+++ b/core-it-support/core-it-plugins/maven-it-plugin-settings/src/main/java/org/apache/maven/plugin/coreit/SettingsReadItMojo.java
@@ -23,7 +23,6 @@ import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.settings.Settings;
 import org.apache.maven.settings.io.xpp3.SettingsXpp3Writer;
-import org.codehaus.plexus.util.IOUtil;
 
 import java.io.File;
 import java.io.FileWriter;
@@ -59,10 +58,8 @@ public class SettingsReadItMojo
             dumpFile.delete();
         }
         dumpFile.getParentFile().mkdirs();
-        FileWriter fw = null;
-        try
+        try ( FileWriter fw = new FileWriter( dumpFile ) )
         {
-            fw = new FileWriter( dumpFile );
             SettingsXpp3Writer writer = new SettingsXpp3Writer();
             writer.write( fw, settings );
         }
@@ -70,11 +67,5 @@ public class SettingsReadItMojo
         {
             throw new MojoExecutionException( e.getMessage(), e );
         }
-        finally
-        {
-            IOUtil.close( fw );
-        }
     }
-
-
 }
--- a/core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
+++ b/core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
@@ -46,6 +46,7 @@ under the License.
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>2.0</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.reporting</groupId>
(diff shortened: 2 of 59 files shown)


core-it-support/core-it-plugins/maven-it-plugin-settings/src/main/java/org/apache/maven/plugin/coreit/SettingsReadItMojo.java
```diff
@@ -59,22 +58,14 @@ public void execute()
...
         }
         catch ( IOException e )
         {
             throw new MojoExecutionException( e.getMessage(), e );
         }
-        finally
-        {
-            IOUtil.close( fw );
-        }
     }
```
psiroky, line 70: These simple changes are to avoid the direct dependency on plexus-utils in this plugin (as it is brought in by one the maven deps, which are now in provided scope - and thus pluxus-utils ends up in provided scope as well.)

core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
```diff
@@ -67,6 +68,40 @@ under the License.
...
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.maven</groupId>
+          <artifactId>maven-settings</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.maven</groupId>
+          <artifactId>maven-settings-builder</artifactId>
+        </exclusion>
+      </exclusions>
```
psiroky, line 104: Not 100% sure about the `exclusion` approach here. Seems to work as the plugin does not need these.

I guess the alternative would be to add these as `provided` dependencies into the POM.

core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
```diff
@@ -132,7 +131,7 @@ static Dependency toDependency( String artifact )
             throws MojoFailureException
     {
         Dependency coordinate = new Dependency();
-        String[] tokens = StringUtils.split( artifact, ":" );
```
psiroky, line 135: Similar as above to avoid adding direct dependency on `plexus-utils`

psiroky:
Once I had these changes done I noticed @slawekjaranowski solved this a bit differently in `maven-enforcer` (https://github.com/apache/maven-enforcer/commit/74228deccd43ec88dadf59fa70d98dfd5ead33c5), by adding the `<scope>provided</scope> into `dependencyManagement`.

For some reason I thought that having `scope` inside `dependencyManagement` is not a good practice (with the exception of BOM `import`) as it should just define the version and then modules using that dependency should decide which scope is appropriate for that specific use case. That being said, in this case it actually makes sense as we want these always provided for the IT plugins. If preferred, I can amend the PR with similar changes leveraging `dependencyManagement`.

core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
slawekjaranowski: or exclusion for `*:*`

psiroky:
Yeah, that is the case in one of the plugins here as well, hence the exclusions.

The `dependencyManagement` solution is quite elegant though as it means way less changes and we can also get rid of the exclusions. I don't have a strong preference at this point, so I will leave the current solution. If someone feels strongly about this I can use the other way with dep management.

core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
psiroky: Hmm, good point. Thanks!

I will try to exclude all the related Maven artifacts with one exclusion.

psiroky:
Hello @slawekjaranowski! Just checking in to see how to proceed with these clean-up PRs (this one and others with deprecated verifier methods). 

Would you have time to take a look at these? Or should I perhaps e-mail to the dev list to ask for reviews?

I don't want to pressure you or anything, just trying to make sure the PRs won't be forgotten. If you plan to take a look in the future and just don't have time right now, that's of course fine.

PR #228, opened 2023-01-11
[MNG-7665] Update Jetty to latest Java8 capable version

cstamas:
Update Jetty to latest 9.4.50, last Java 8 capable version. Apply API changes and also remove jetty-servlet as ITs does NOT use servlet aspect of Jetty HTTP server.. This will simplify our lives in future, to make jump to later versions (that are Java 11+). Also, drop jetty-servlet as test do not use any servlet aspect.

---

https://issues.apache.org/jira/browse/MNG-7665

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4470AuthenticatedDeploymentToProxyTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4470AuthenticatedDeploymentToProxyTest.java
@@ -43,15 +43,12 @@ import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 import org.eclipse.jetty.server.handler.HandlerList;
-import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.util.security.Constraint;
 import org.eclipse.jetty.util.security.Password;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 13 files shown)


gnodet (approved):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4470AuthenticatedDeploymentToProxyTest.java
```diff
@@ -92,7 +89,7 @@ public void handle( String target, Request baseRequest, HttpServletRequest reque
                 if ( auth != null )
                 {
                     auth = auth.substring( auth.indexOf( ' ' ) + 1 ).trim();
-                    auth = B64Code.decode( auth, StandardCharsets.US_ASCII.name() );
+                    auth = new String( java.util.Base64.getDecoder().decode( auth ), StandardCharsets.US_ASCII );
```
slawekjaranowski, line 92: Why not `import java.util.Base64`?

cstamas: no special reason, I believe at one point both classes were present (Jetty deprecated B64code and j.u.Base64) and my IDE just choose to fully name it? Should I add it to import?

michael-o (approved):

slawekjaranowski (approved):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4470AuthenticatedDeploymentToProxyTest.java
slawekjaranowski: 😄

PR #230, opened 2023-01-12
[MNG-7667] Fix Bootstrap IT

cstamas:
This change extends Bootstrap IT (1st IT that fills in local repo) to hunt down some plugins that are bound in lifecycles provided by Maven under test.

Changes:
* the Bootstrap IT invokes only the phase where DownloadMojo is bound
* DownloadMojo (the "bootstrap") changed to do "this or that": it either resolves all artifacts from provided file OR collects plugins for given project packaging plugin versions provided by tested Maven default lifecycle.
* Split more Bootstrap project POMs, now we have 4 modules, we can add more (even extension if needed)

---

https://issues.apache.org/jira/browse/MNG-7667

### DIFF
--- a/core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
+++ b/core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
@@ -31,6 +31,7 @@ import org.apache.maven.RepositoryUtils;
 import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Component;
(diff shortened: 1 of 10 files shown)


core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
```diff
@@ -104,6 +105,21 @@ public void execute() throws MojoFailureException
             }
         }
 
+        // HACK: this plugin is executing in bootstrap project that has packaging=pom, but still enlists install/deploy
+        try
+        {
+            for ( Plugin plugin : session.getCurrentProject().getBuildPlugins() )
+            {
+                dependencies.add( toDependency(
```
slawekjaranowski, line 113: What about  make `dependencies`  as a set to avoid duplicates?

cstamas: Resolving when once done is fast: basically presence of file(s) on disk happens, so duplication (for now) is not something I'd worry about

gnodet (approved):

core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
```diff
@@ -82,8 +86,10 @@
     @Override
     public void execute() throws MojoFailureException
     {
+        // this or that: either resolver file listed artifacts or collect lifecycle packaging plugins
         if ( file != null && file.exists() )
         {
+            System.out.println( "Collecting artifacts from file: " + file );
```
slawekjaranowski, line 92: Why system out instead of Mojo log?

cstamas: I just followed existing code, see below, in this mojo nothing uses logger, but everything uses System.out

slawekjaranowski (approved):

PR #231, opened 2023-01-12
Fix MavenITmng5669ReadPomsOnce

slawekjaranowski:
In test, we check if pom files are read only once.

When we update plugin or dependencies in test,
number of read items can change again.

So should be good enough to check if duplicate items exist in log file.

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5669ReadPomsOnce.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng5669ReadPomsOnce.java
@@ -19,17 +19,16 @@ package org.apache.maven.it;
  * under the License.
  */
 
-import org.apache.maven.shared.verifier.util.ResourceExtractor;
-import org.apache.maven.shared.verifier.Verifier;
-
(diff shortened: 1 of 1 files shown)


core-it-suite/src/test/resources-filtered/bootstrap.txt
```diff
@@ -79,13 +79,15 @@ org.apache.maven.plugins:maven-dependency-plugin:3.3.0
 org.apache.maven.plugins:maven-deploy-plugin:${stubPluginVersion}
 org.apache.maven.plugins:maven-deploy-plugin:2.7
 org.apache.maven.plugins:maven-deploy-plugin:3.0.0
+org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M2
```
cstamas, line 82: these will not be needed once https://github.com/apache/maven-integration-testing/pull/230 merged

slawekjaranowski: rebased

gnodet (approved):

cstamas (approved):

PR #232, opened 2023-01-13
Remove a few system properties before initializing the maven verifier

gnodet:


gnodet:
The `maven.conf` is only used as a default in the `m2.conf`, so when using a non forked verifier, the current system property is set. It points to the maven home from the running maven and not the one that is tested.

core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
```diff
@@ -50,6 +50,9 @@ private static void infoProperty( PrintStream info, String property )
     {
         try
         {
+            System.clearProperty( "maven.bootclasspath" );
+            System.clearProperty( "maven.conf" );
+            System.clearProperty( "classworlds.conf" );
```
slawekjaranowski, line 55: `maven.bootclasspath` and `classworlds.conf` are used by Verifier - can be null, will be discovered according to `maven.home` used by test. 
By the way, should by fixed in Verifier 😄 I have a plan for it ...

`maven.conf` is used by Maven itself ... grrry 
https://github.com/apache/maven/search?l=Java&q=maven.conf

I hope that `DefaultMavenSettingsBuilder` is not used ....

slawekjaranowski (approved):

michael-o (commented):
I consider this wrong because they should be at most managed from within Verifier. Those are *system* properties and when these are required then a *forked* verifier must be used.

slawekjaranowski:
> I consider this wrong because they should be at most managed from within Verifier. Those are _system_ properties and when these are required then a _forked_ verifier must be used.

You are right, should be manage by Verifier, I checked how properties are used in Verifier ... 
also plexus launcher configure method is called by Verifier  so props are needed by plexus

gnodet:
> I consider this wrong because they should be at most managed from within Verifier. Those are _system_ properties and when these are required then a _forked_ verifier must be used.

I don't really see how that's different from https://github.com/apache/maven-integration-testing/pull/232

Another possibility is to have the Verifier set this property explicitly to the value `${maven.home}/conf`.  And if you think it's wrong that the Verifier change these properties, then let's do it in this PR as this won't affect anyone else.

But really, having the tests not being isolated is a bad idea.

michael-o:
> > I consider this wrong because they should be at most managed from within Verifier. Those are _system_ properties and when these are required then a _forked_ verifier must be used.
> 
> I don't really see how that's different from #232
> 
> Another possibility is to have the Verifier set this property explicitly to the value `${maven.home}/conf`. And if you think it's wrong that the Verifier change these properties, then let's do it in this PR as this won't affect anyone else.
> 
> But really, having the tests not being isolated is a bad idea.

I am not isolation, but isolation needs to happen by the framework executing the tests and if necessary, create a new process for it.

gnodet:
> > > I consider this wrong because they should be at most managed from within Verifier. Those are _system_ properties and when these are required then a _forked_ verifier must be used.
> > 
> > 
> > I don't really see how that's different from #232
> > Another possibility is to have the Verifier set this property explicitly to the value `${maven.home}/conf`. And if you think it's wrong that the Verifier change these properties, then let's do it in this PR as this won't affect anyone else.
> > But really, having the tests not being isolated is a bad idea.
> 
> I am not isolation, but isolation needs to happen by the framework executing the tests and if necessary, create a new process for it.

Sorry, bad link above.  I meant https://github.com/apache/maven-verifier/blob/70789f300e3bb30dd3d57f4d81fbdeee7b0310ee/src/main/java/org/apache/maven/shared/verifier/Embedded3xLauncher.java#L202-L206

So, as @slawekjaranowski indicated, the two properties `maven.bootclasspath` and `classworlds.conf` are actually input for the Verifier, so that's part of the verifier's configuration, so they do belong here imho.  Unless Verifier is changed somehow, but at this point, I think it makes sense.

The third one, the `maven.conf` should be cleared by the verifier, but it's also an input. So it's a possible valid use case to actually set the `maven.conf`, so again, it can't really always be cleared by the verifier.

I think the problem is that the verifier should not use system properties to configure itself and should be given a `Map<String, String>` from where it can find various configuration bits to isolate the call to the verifier from the environment.  But that's a bigger change, so until the Verifier is improved and change the way it's configured, I think this PR is a good workaround @michael-o

michael-o:
> > > > I consider this wrong because they should be at most managed from within Verifier. Those are _system_ properties and when these are required then a _forked_ verifier must be used.
> > > 
> > > 
> > > I don't really see how that's different from #232
> > > Another possibility is to have the Verifier set this property explicitly to the value `${maven.home}/conf`. And if you think it's wrong that the Verifier change these properties, then let's do it in this PR as this won't affect anyone else.
> > > But really, having the tests not being isolated is a bad idea.
> > 
> > 
> > I am not isolation, but isolation needs to happen by the framework executing the tests and if necessary, create a new process for it.
> 
> Sorry, bad link above. I meant https://github.com/apache/maven-verifier/blob/70789f300e3bb30dd3d57f4d81fbdeee7b0310ee/src/main/java/org/apache/maven/shared/verifier/Embedded3xLauncher.java#L202-L206
> 
> So, as @slawekjaranowski indicated, the two properties `maven.bootclasspath` and `classworlds.conf` are actually input for the Verifier, so that's part of the verifier's configuration, so they do belong here imho. Unless Verifier is changed somehow, but at this point, I think it makes sense.
> 
> The third one, the `maven.conf` should be cleared by the verifier, but it's also an input. So it's a possible valid use case to actually set the `maven.conf`, so again, it can't really always be cleared by the verifier.
> 
> I think the problem is that the verifier should not use system properties to configure itself and should be given a `Map<String, String>` from where it can find various configuration bits to isolate the call to the verifier from the environment. But that's a bigger change, so until the Verifier is improved and change the way it's configured, I think this PR is a good workaround @michael-o

Then please mark those three lines as a workaround for the time being.

michael-o (approved):

PR #234, opened 2023-01-14
[MNG-7629] Change reactor reader to copy packaged artifacts and reuse them across builds if needed

gnodet:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3043BestEffortReactorResolutionTest.java
```diff
@@ -25,6 +25,7 @@
 import java.io.File;
 import java.util.List;
 
+import org.junit.jupiter.api.Disabled;
```
mthmulders, line 28: I don't think this import is used?

mthmulders (approved):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng3043BestEffortReactorResolutionTest.java
gnodet: Good catch, thx !

PR #236, opened 2023-01-16
[MNG-7667] Fix DownloadMojo to properly resolve

cstamas:
The m-p-p depends on m-p-tools-ant and m-p-tools-beanshell as runtime/optional, and the way DownloadMojo resolves every artifact causes that OptionalDependencySelector filter these out even at "collection" stage (as depth >= 2).

Hence, fix DownloadMojo to resolve properly, that actually shifts all the tree "one level up", and hence result is same as in Maven, as the optional dependencies become "first level sibling".

---

https://issues.apache.org/jira/browse/MNG-7667


cstamas:
Am redoing this, instead of hack make it proper, turns out reading POM is not such a big overhead (and is anyway done already as part of collection).

PR #237, opened 2023-01-17
[MNG-7608] Adjust IT for default transport

cstamas:
If 3.9 uses native as default as well, there is no need anymore to distinguish between 3.9 and 4+.

---

https://issues.apache.org/jira/browse/MNG-7608

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7470ResolverTransportTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7470ResolverTransportTest.java
@@ -13,7 +13,7 @@ import org.junit.jupiter.api.Test;
 
 /**
  * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-7470">MNG-7470</a>:
- * check that Maven two bundled transports works as expected.
+ * check that Maven bundled transports work as expected.
  */
 public class MavenITmng7470ResolverTransportTest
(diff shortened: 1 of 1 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7470ResolverTransportTest.java
```diff
@@ -105,14 +105,7 @@ private void performTest( /* nullable */ final String transport, final String lo
     public void testResolverTransportDefault()
             throws Exception
     {
-        if ( matchesVersionRange( "[4.0.0-alpha-3,)" ) )
-        {
-            performTest( null, NATIVE_LOG_SNIPPET );
-        }
-        else // [3.9.0,) but not 4, see above and ctor
-        {
-            performTest( null, WAGON_LOG_SNIPPET );
```
michael-o, line 114: Can this snippet be deleted then?

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7470ResolverTransportTest.java
cstamas: No, as this class has 3 tests: default, forced-wagon, forced-native and is used in forced-wagon test. This change is only about default one (nothing forced).

PR #238, opened 2023-01-19
[MNG-7661] Fix few build / compiler warnings

psiroky:
Just couple of minor fixes. I realized I had these stashed away for some time and just forgot to create PR earlier


core-it-support/core-it-javaagent/pom.xml
```diff
@@ -53,6 +53,17 @@ under the License.
...
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>module-info.class</exclude>
+                    <exclude>META-INF/MANIFEST.MF</exclude>
+                  </excludes>
+                </filter>
+              </filters>
+            </configuration>
```
psiroky, line 66: "Fixes" these warnings:
```
[INFO] --- maven-shade-plugin:3.4.1:shade (default) @ core-it-javaagent ---
[INFO] Including org.ow2.asm:asm:jar:7.3.1 in the shaded jar.
[INFO] Including org.ow2.asm:asm-commons:jar:7.3.1 in the shaded jar.
[INFO] Including org.ow2.asm:asm-tree:jar:7.3.1 in the shaded jar.
[INFO] Including org.ow2.asm:asm-analysis:jar:7.3.1 in the shaded jar.
[INFO] Including org.ow2.asm:asm-util:jar:7.3.1 in the shaded jar.
[INFO] Dependency-reduced POM written at: /home/psiroky/code/maven-ecosystem/maven-integration-testing/core-it-support/core-it-javaagent/dependency-reduced-pom.xml
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] asm-7.3.1.jar, asm-analysis-7.3.1.jar, asm-commons-7.3.1.jar, asm-tree-7.3.1.jar, asm-util-7.3.1.jar, core-it-javaagent-2.1-SNAPSHOT.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See https://maven.apache.org/plugins/maven-shade-plugin/
```

core-it-support/core-it-plugins/maven-it-plugin-class-loader/maven-it-plugin-class-loader/pom.xml
```diff
@@ -44,6 +44,10 @@ under the License.
   </description>
   <inceptionYear>2008</inceptionYear>
 
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
```
psiroky, line 49: This project is not using the parent (since it apparently cannot as described in the comment), so some of the config needs to be duplicated.

core-it-support/core-it-plugins/maven-it-plugin-site/pom.xml
```diff
@@ -72,6 +72,12 @@ under the License.
       <groupId>org.apache.maven.reporting</groupId>
       <artifactId>maven-reporting-exec</artifactId>
       <version>1.6.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.maven</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
```
psiroky, line 80: This has been already done as part of https://github.com/apache/maven-integration-testing/pull/223, but then I think it got accidentally reverted when merging some other changes.

core-it-support/maven-it-plugin-bootstrap/src/main/java/org/apache/maven/its/bootstrap/DownloadMojo.java
```diff
@@ -74,7 +74,7 @@
     @Component
     private RepositorySystem repositorySystem;
 
-    @Component
+    @Parameter( defaultValue = "${session}", readonly = true )
```
psiroky, line 77: This is suggested by `maven-plugin-plugin`
```
INFO] --- maven-plugin-plugin:3.6.4:descriptor (default-descriptor) @ maven-it-plugin-bootstrap ---
[INFO] Using 'UTF-8' encoding to read mojo source files.
[WARNING] Deprecated @Component annotation for 'session' field in org.apache.maven.its.bootstrap.DownloadMojo: replace with @Parameter( defaultValue = "${session}", readonly = true )
```

core-it-support/core-it-javaagent/pom.xml
```diff
@@ -53,6 +53,17 @@ under the License.
             <goals>
               <goal>shade</goal>
             </goals>
+            <configuration>
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>module-info.class</exclude>
+                    <exclude>META-INF/MANIFEST.MF</exclude>
```
slawekjaranowski, line 62: m-jar-p adds items to MANIFEST - see line 40-43 - we need its, so we don't drop MANIFEST
we can try https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html#ManifestResourceTransformer

psiroky: Nice catch, thanks! I would have hoped that "something" would fail the build if those manifest entries are missing -- seeing the linked PR, I assume this is only failing with Maven 4.x, which is currently not part of PR checks.

In any case, I adjusted the filter to exclude only the `asm` artifacts, which should fix the issue (the build still won't pass with Maven 4.x, but the reported errors/failures seem to be consistent with the ones I am seeing on master branch)

slawekjaranowski:
@psiroky thanks

PR #240, opened 2023-02-14
Run MNG-7160 test also for Maven 3.8.8+

famod:
For https://github.com/apache/maven/pull/999


michael-o (changes requested):
The entire test needs to be back ported to the `maven-3.8.x` branch. Master is for 4 and 3.9.x only.

famod:
Ah, my bad - I missed that somehow. I ran it locally like that and it passed.

Anyway, I'll create another PR then.

michael-o:
> Ah, my bad - I missed that somehow. I ran it locally like that and it passed.
> 
> Anyway, I'll create another PR then.

Not necessary, you can rebase, force push and change the merge base here.

famod:
Turns out the test is already there in the 3.8.x branch.

michael-o:
Change applied.

PR #241, opened 2023-02-15
No logic change, just improve logging

cstamas:
This IT helped a lot to figure out problems around parallel deploy, and I had locally applied these changes to improve ability to track down what is actually happening on server side.


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4470AuthenticatedDeploymentToProxyTest.java
```diff
@@ -84,15 +84,16 @@ protected void setUp()
             public void handle( String target, Request baseRequest, HttpServletRequest request,
                                 HttpServletResponse response )
             {
-                System.out.println( "Handling " + request.getMethod() + " " + request.getRequestURL() );
+                String tid = Thread.currentThread().getName();
+                System.out.println( tid + " Handling (proxy) " + request.getMethod() + " " + request.getRequestURL() );
```
michael-o, line 88: That is not the tid, that is the name.

michael-o (changes requested):

slawekjaranowski (approved):

PR #242, opened 2023-02-17
Use Maven 3.9.4 on GitHub

slawekjaranowski:


michael-o (changes requested):
I wouldn't do this before at least 3.9.1, .0 contains too many open issues for now.

slawekjaranowski:
I would like to go a step forward and also add 4.x 
After review of #238 I found issue with 4.x
Simply we should have a possibility to detect an issue early

What do you think?

slachiewicz (approved):

mthmulders:
How about a matrix, with Maven 3.9.x and 4.0.0-alpha-x (where `x` is latest for both)?

slawekjaranowski:
> How about a matrix, with Maven 3.9.x and 4.0.0-alpha-x (where `x` is latest for both)?

I'm not sure about matrix build here - it should be a simple test for changing in repo.
We execute ITs when Maven build itself for specific change.

slawekjaranowski:
> I wouldn't do this before at least 3.9.1, .0 contains too many open issues for now.

I hope that 3.9.4 is better than 3.9.1 😄

michael-o (approved):

cstamas (approved):

PR #244, opened 2023-02-20
[MNG-7679] Single Mojo no POM -> NPE

cstamas:
IT for this use case, Maven 3.9.0 suffers from this, is excluded. I guess some Maven 4.x suffers as well?

---

https://issues.apache.org/jira/browse/MNG-7679


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7679SingleMojoNoPomTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * Executing single mojo without POM should not NPE.
+ *
+ */
+class MavenITmng7679SingleMojoNoPomTest
+    extends AbstractMavenIntegrationTestCase
+{
+
+    public MavenITmng7679SingleMojoNoPomTest()
+    {
+        super( "(,3.9.0),(3.9.0,)" );
```
michael-o, line 39: `[3.8.8,)` All three are broken and need fíxing

michael-o (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7679SingleMojoNoPomTest.java
cstamas: Let's enumerate which version(s) to leave out:
* 3.8.7
* 3.9.0
* 4.0.0-alpha4
?

michael-o: Ah, right!

`[3,8.8,3.9.0),[3.9.1,4.0.0-alpha-4),[4.0.0-alpha-5,)`?

cstamas:
Pushed what I have locally, but locally still skips 3.9.1-SNAPSHOT :disappointed: 
@michael-o pls look

PR #247, opened 2023-02-24
Detecting tests which was not added to TestSuiteOrdering

slawekjaranowski:
Even test is skipped by `@Disabled` annotation
newer surefire schedule it for execution.

When tests are missing in TestSuiteOrdering
are executed as one of first tests in unpredictable order.

We have many options:
 - left as is
 - show info as in PR
 - break a build

### DIFF
--- a/core-it-suite/pom.xml
+++ b/core-it-suite/pom.xml
@@ -498,6 +498,8 @@ under the License.
           <includes>
             <include>**/MavenIT*.java</include>
           </includes>
+          <!-- test annotated by @Tag("disabled") will be skipped from executions -->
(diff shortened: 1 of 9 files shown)


michael-o:
I don't understand why they aren't skipped by default.

slawekjaranowski:
> I don't understand why they aren't skipped by default.

Are skipped, really executed with skipped status. 
But are not ordered. 
With older version of surefire was not executed / or reported, was invisible.

michael-o (approved):

cstamas (approved):

PR #248, opened 2023-02-24
MavenITmng5669ReadPomsOnce and BF collector

slawekjaranowski:
Test with BF fails

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -49,7 +49,7 @@ jobs:
           cache: 'maven'
 
       - name: Set up Maven
-        run: mvn --errors --batch-mode --show-version wrapper:wrapper "-Dmaven=3.8.7"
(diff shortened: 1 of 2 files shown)


cstamas (approved):

.github/workflows/maven-verify.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           cache: 'maven'
 
       - name: Set up Maven
-        run: mvn --errors --batch-mode --show-version wrapper:wrapper "-Dmaven=3.8.7"
+        run: mvn --errors --batch-mode --show-version wrapper:wrapper "-Dmaven=3.9.0"
```
michael-o, line 52: Is this the version which is used to run the actual ITs or just compile the project?

slawekjaranowski: in this build for both - but ok test is only for 4.x

PR #249, opened 2023-03-02
[MNG-5181] Improve "missing dependency" message

cstamas:
Related IT change, as the change for this issue made MNG-7128 IT to fail, as it was too eagerly matching whole error message string that has now due MNG-5181 changed.

To fix it, match only the key parts of the expected error message, that project failed, due transfer failure, as insecure mirror was blocked. This makes IT assert same thing, but is more flexible regarding error text change in MNG-5181.

---

Context: the modified IT fails in this PR https://github.com/apache/maven/pull/1025 as Resolver 1.9.6 carries extra information about artifact availability in the message, that makes the IT full string match fail now.

---

https://issues.apache.org/jira/browse/MNG-5185


hboutemy (approved):
LGTM

PR #252, opened 2023-03-13
[MNG-7737] IT for profiles

cstamas:
Just a simple IT that would catch issues like Maven 3.9.0 went out with, where files were wrongly interpolated: MNG-7726

---

https://issues.apache.org/jira/browse/MNG-7737

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7737ProfileActivationTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7737ProfileActivationTest.java
@@ -1,0 +1,101 @@
+package org.apache.maven.it;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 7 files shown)


cstamas:
Q: shouldn't the POM "pom-active-default" become active as well?

slawekjaranowski:
> Q: shouldn't the POM "pom-active-default" become active as well?

As I know - no, activeByDefault - is only activated when no other profiles are active

From docs:

> This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.

cstamas:
Oh, but then why is "settings-active-default" active?

cstamas:
Seems this is true for POM profiles only: "pom-active-default" gets enabled ONLY when I comment out any other profile in POM. The presence of ANY settings (default or other way activated profile) does not matter.

cstamas:
Verified locally that IT fails w/ 3.9.0 and passes with 3.9.1 (for this the IT range needs change)

PR #260, opened 2023-05-10
[MNG-7740] Add test for target directory being flooded with consumer*pom files

Giovds:
For [this PR](https://github.com/apache/maven/pull/1105)

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest.java
@@ -36,7 +36,7 @@ class MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest
 
     protected MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest()
     {
-        super( "[4.0.0-alpha-5,)" );
+        super( "[4.0.0-alpha-6,)" );
     }
 
     @Test
@@ -49,7 +49,6 @@ class MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest
         verifier.addCliArgument( "validate" );
 
         verifier.execute();
-        verifier.execute();
 
         verifier.verifyErrorFreeLog();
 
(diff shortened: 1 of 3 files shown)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+class MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest
+        extends AbstractMavenIntegrationTestCase
+{
+
+    protected MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest()
+    {
+        super( "[4.0.0-alpha-5,)" );
```
mthmulders, line 39: I think by now this should be alpha-6.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest.java
```diff
@@ -0,0 +1,63 @@
...
+        verifier.execute();
+        verifier.execute();
+
+        verifier.verifyErrorFreeLog();
+
+        try(Stream<Path> stream = Files.walk(testDir.toPath())) {
+            final List<Path> consumerFiles = stream.filter( path -> path.getFileName().toString().contains("consumer")
+                            && path.getFileName().toString().contains( "pom" ))
+                    .collect( Collectors.toList());
+            assertTrue("Expected only 1 consumer pom file.", consumerFiles.size() == 1);
```
mthmulders, line 60: If we end up with the solution that cleans files on Maven exit (https://github.com/apache/maven/pull/1117) this should check for 0 rather than 1.

mthmulders (changes requested):

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7740ConsumerBuildShouldCleanUpOldFilesTest.java
Giovds: I've also removed the second execute of `verifier.execute()` as there is no need to generate an initial file to validate the old one is deleted.

mthmulders:
I doesn't make to much sense to run it right now, as it will test against Maven's master branch which does not yet have https://github.com/apache/maven/pull/1117 merged. @gnodet, agreed to merge this PR once either @Giovds or your PR for removing those consumer*pom files is merged into master?

PR #261, opened 2023-05-19
[MNG-7788] Fix for IT MavenITmng6562WarnDefaultBindings

cstamas:
The MavenITmng6562WarnDefaultBindings IT tests new Maven4 feature "fail on severity" (for WARN level).

The stub core plugin maven-resources-plugin depends on maven-compat, and it emits plugin validation warning.

BEFORE MNG-7788 was merged, the warnings were logged AFTER build summary and was not picked up by "fail on severity".

AFTER MNG-7788 merged, it causes that warnings breaks the build, as "fail on severity" is set to WARN level.

Fix:
The stub maven-resources-plugin does NOT need to depend on maven-compat, what it really needs is maven-core (for MavenProject).


michael-o (approved):
Good explanation.

PR #263, opened 2023-05-24
[MNG-7740] Remove old temporary consumer*pom files from buildDir

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7740
IT for https://github.com/apache/maven/pull/1117


gnodet:
Supersedes #260 
The repository + branch name must match so that they can be tested together

mthmulders (approved):
As a general remark, we don't apply the new coding style for the IT's?

gnodet:
> As a general remark, we don't apply the new coding style for the IT's?

I think that's because we haven't upgraded the parent to a recent version.  We definitely should.

PR #268, opened 2023-06-05
[MNG-7804] IT for plugin execution order

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MNG-7804
Main PR: https://github.com/apache/maven/pull/1147


ctubbsii (commented):
Good test coverage. Seems like a nice feature addition. Still some outstanding questions, though:

* Does this require the POM/4.2.0 model? Which versions of Maven can use that POM version?
* How does this work with priorities defined in a parent POM or profile?
* What is the numerical type of the priority? Is it an integer, decimal, long, non-negative, strictly positive?
* What if two executions have the same priority? Is that an error, or is it arbitrarily ordered?

PR #271, opened 2023-06-19
[MNG-7819] Add IT that excercise the bug

cstamas:
Template/reproducer: https://gist.github.com/cstamas/3f15775bd4785a5f4b157399f52af6fc

While local repo is empty.

---

https://issues.apache.org/jira/browse/MNG-7819

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7819FileLockingWithSnapshotsTest.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7819FileLockingWithSnapshotsTest.java
@@ -41,7 +41,8 @@ class MavenITmng7819FileLockingWithSnapshotsTest extends AbstractMavenIntegratio
     private int port;
 
     protected MavenITmng7819FileLockingWithSnapshotsTest() {
-        super("[3.9.0,)");
+        // broken: maven 3.9.2 and 4.0.0-alpha-5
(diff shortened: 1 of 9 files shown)


cstamas:
Adding this IT should be safe: it will fail with 3.9.2 and alpha-5, but it also excludes, and starting with 3.9.3 it should run

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7819FileLockingWithSnapshotsTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+class MavenITmng7819FileLockingWithSnapshotsTest extends AbstractMavenIntegrationTestCase {
+
+    private Server server;
+
+    private int port;
+
+    protected MavenITmng7819FileLockingWithSnapshotsTest() {
+        // broken: maven 3.9.2 and 4.0.0-alpha-5
+        super("[3.9.0,3.9.2),(3.9.2,],[4.0.0-alpha-6,)");
```
cstamas, line 45: [3.9.0,3.9.2),(3.9.2,4.0.0-alpha-5),[4.0.0-alpha-6,) instead?

cstamas: bah, no, would skip 3.9.3...

gnodet (commented):
Please fix the range to include `3.9.3`, otherwise fine.

PR #295, opened 2023-09-20
[MNG-7891] add an example for a configurable extension

laeubi:
This adds a simple extension that prints out a message at session start and after projects are read.

There is also an example consumer of that extension that configures the message to print the start of the build and the value of a user supplied value on the commandline.


laeubi:
With stock maven this currently produces:
```
[ERROR] Error executing Maven.
[ERROR] Unrecognised tag: 'configuration' (position: START_TAG seen ...</version>\n\t\t<configuration>... @6:18) 
```

I would expect instead that it prints two messages when started with `mvn install -DuserValue=HelloWorld`:

```
The session has just started at <the maven build start timestamp>
All projects are read now! The build will start and the user has passed -DuserValue=HelloWorld on the commandline
```

gnodet:
Superseded with #296

PR #298, opened 2023-09-23
Use 1.8 for source/target in ITs - allow to execute by newer JDK

slawekjaranowski:
1.7 was removed in JDK 20


slachiewicz (commented):
Yes, time to move to Java 8 as baseline also for Maven ITs

slachiewicz (approved):

PR #299, opened 2023-09-24
[MNG-7895] Support ${project.basedir} in file profile activation - fix ITs

slawekjaranowski:


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7836AlternativePomSyntaxTest.java
```diff
@@ -27,8 +27,11 @@
 
 import org.apache.maven.shared.verifier.Verifier;
 import org.apache.maven.shared.verifier.util.ResourceExtractor;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+// TODO - Temporarily disabled - until test will be fixed
+@Disabled
```
slawekjaranowski, line 34: @gnodet - do you work on fixing it? 
If not I can see on it.

PR #308, opened 2023-09-27
Fix expectations on IT MNG-7836 following the modifications on the consumer pom

gnodet:


slawekjaranowski:
Can we change in: https://github.com/apache/maven-integration-testing/pull/307

slawekjaranowski (approved):

PR #311, opened 2023-10-13
[MNG-7772] Test for extensions.xml defined in an ext directory in the…

juulhobert:
Integration test for [PR](https://github.com/apache/maven/pull/1280) and [Jira issue](https://issues.apache.org/jira/browse/MNG-7772)


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * This is a test set for <a href="https://issues.apache.org/jira/browse/MNG-7772">MNG-7772</a>:
+ * check that Maven fails if it cannot load core extensions contributed by <code>${user.home}/.m2/extensions.xml</code>.
+ */
+public class MavenITmng7772CoreExtensionsNotFoundTest extends AbstractMavenIntegrationTestCase {
+
+    public MavenITmng7772CoreExtensionsNotFoundTest() {
+        super("[3.9.0,)");
```
mthmulders, line 40: Since your change on Maven itself is on the **master** branch, I'd expect the version range to start with "4.0.0-alpha-8".

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Test
+    public void testCoreExtensionsNotFound() throws Exception {
+        File testDir = ResourceExtractor.simpleExtractResources(getClass(), "/mng-7772-core-extensions-not-found");
+
+        Path m2TestDirPath = Paths.get(testDir.toPath().toString(), ".m2");
+        Files.createDirectories(Paths.get(System.getProperty("user.home"), ".m2"));
```
mthmulders, line 48: I'm not 100% sure this is the location of the Maven user directory in the Core Integration Suite. I hope someone else knows for sure....

mthmulders (changes requested):

mthmulders:
I'd love to also see a "positive" test case, e.g., one that demonstrates that Maven indeed activates an extension that is installed in the user-level Maven directory. I think that test case should fork to a new JVM to prevent the extension leaking into other test cases.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
slawekjaranowski: We can not make changes in user directory by executing test

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
```diff
@@ -0,0 +1,81 @@
...
+        // Test cannot run when there is already an extensions.xml file present in the ${user.home}/.m2 folder
+        assertFalse(Files.exists(to));
+
+        try {
+            Files.copy(m2TestDirPath.resolve("extensions.xml"), to);
+            Verifier verifier = newVerifier(testDir.getAbsolutePath());
+
+            try {
+                verifier.addCliArgument("validate");
+                verifier.execute();
```
slawekjaranowski, line 60: I would like to try:

```
        verifier.setSystemProperty("user.home", mocUserHom_with_m2 );

```

slawekjaranowski: yes I know it is not the best ... but better than interact with real user directory

slawekjaranowski (approved):

slawekjaranowski (changes requested):
.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
juulhobert: Changed it to `"(4.0.0-alpha-7,)"`

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7772CoreExtensionsNotFoundTest.java
juulhobert: I see why it's not preferable to use the real user home. Somehow `.setSystemProperty("user.home"` was not working but I found that test MavenITmng0553SettingsAuthzEncryptionTest already had a method for setting the user home. I've reused that code and that's working :)

core-it-suite/src/test/java/org/apache/maven/it/ItUtils.java
```diff
@@ -56,4 +63,24 @@ public static String calcHash(File file, String algo) throws Exception {
...
+        setUserHome(verifier, file.toPath());
+    }
+
+    public static void setUserHome(Verifier verifier, Path home) {
+        // NOTE: We set the user.home directory instead of say settings.security to reflect Maven's normal behavior
+        String path = home.toAbsolutePath().toString();
+        if (path.indexOf(' ') < 0) {
+            verifier.setEnvironmentVariable("MAVEN_OPTS", "-Duser.home=" + path);
+        } else {
+            verifier.setEnvironmentVariable("MAVEN_OPTS", "\"-Duser.home=" + path + "\"");
```
slawekjaranowski, line 83: We also need here 
```
        verifier.setForkJvm(true);
```
environment variable can be set only in fork mode of verifier

juulhobert: It's not trivial to get the forkJvm value, therefore i've only added a comment to warn future developers about it. I hope you're ok with it.

core-it-suite/src/test/resources/mng-7772-core-extensions-lib-ext/extension/.gitignore
mthmulders, line 1: Nit: this wasn't covered by existing **.gitignore** rules?

core-it-suite/src/test/resources/mng-7772-core-extensions-found/extension/.gitignore
mthmulders, line 1: Nit: this wasn't covered by existing **.gitignore** rules?

mthmulders (approved):

mthmulders:
I've kicked off the GitHub Action build for the corresponding PR in Maven Core. If those changes build correctly, it should trigger these new tests.

mthmulders:
The two extensions are completely identical. Can't we reuse them?

Also, in the testing of [MNG-7772](https://github.com/apache/maven/pull/1280), building the Extension fails somehow. I'm trying to reproduce locally but don't yet understand why...

mthmulders (approved):
Nice work, thanks for providing these extensive test cases!

juulhobert:
> Nice work, thanks for providing these extensive test cases!

Thanks

slawekjaranowski (approved):

PR #317, opened 2023-11-07
[MNG-7909] Apply Resolver 2.0.0 changes

cstamas:
Changes:
* IT MNG-4368 "smart install" undone
* IT MNG-5175 force Wagon as it uses Wagon config
* IT MNG-7470 adapt to new default and new transport


gnodet (approved):

gnodet:
You need to make your changes conditional to the maven version I suppose.

gnodet:
> I had all the assumptions inverted :) This should make it, also applied some cleanup

Yeah, much better now.  The assumptions are much easier to read and the test actually works.

PR #324, opened 2023-12-14
[MNG-7967] Verify javadoc:jar is invoked

cstamas:
It uses javadoc:jar goal from locked-down version of Javadoc plugin: Bug is that it refuses to run as 4-alpha-9 returns here language="none". If it would run, it would explode, as project contains java source with intentionally broken javadoc.

---

https://issues.apache.org/jira/browse/MNG-7967


gnodet:
Why choosing a bad javadoc ? Isn't the purpose of the test to check if javadoc was run ? Why do we care about a bad javadoc failure here ?

cstamas:
Bad javadoc is the "honeypot", as if javadoc fails we knew it ran... otherwise (and with bug) we just get successful build but in reality goal did not run.

gnodet (approved):

PR #330, opened 2024-01-01
Enables debug logs in MavenITmng6127

CrazyHZM:
I can't reproduce the unstable 6127 IT in the local environment. With the debug log enabled, I will try to troubleshoot the problem through the debug log.


slawekjaranowski (approved):

slawekjaranowski (commented):
We have two execution on Maven here, both use the same log file.
We can also set different logs for those.

`verifier.setLogFileName `

CrazyHZM:
> We have two execution on Maven here, both use the same log file. We can also set different logs for those.
> 
> `verifier.setLogFileName `

I have found the log from https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven/job/master/509/artifact/core-it-suite-logs-linux-jdk11.zip.

slawekjaranowski:
ok, we have different logs in project and plugin directory

PR #332, opened 2024-01-11
[MNG-8005] Add test for ide WorkspaceReader

jonasrutishauser:


cstamas (approved):
LGTM, I am really surprised we did not have IT for this...

Facts:
* This PR with Maven 3.8.x (the PR is built with it) PASS
* Locally verified that current master of Maven FAILS
* Locally verified that https://github.com/apache/maven/pull/1385 makes master PASS

PR #333, opened 2024-03-29
[MNG-7344] Integration test for effective pom with imports

juulhobert:
Main PR: https://github.com/apache/maven/pull/603

### DIFF
--- a/core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
+++ b/core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
@@ -121,6 +121,9 @@ public class TestSuiteOrdering implements ClassOrderer {
          * a fail fast technique as well.
          */
         suite.addTestSuite(MavenITmng7344EffectivePomPluginTest.class);
+        suite.addTestSuite(MavenITmng7255InferredGroupIdTest.class);
+        suite.addTestSuite(MavenITmng8220ExtensionWithDITest.class);
+        suite.addTestSuite(MavenITmng8181CentralRepoTest.class);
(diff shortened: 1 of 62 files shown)


core-it-suite/src/test/java/org/apache/maven/it/TestSuiteOrdering.java
```diff
@@ -777,6 +777,7 @@ public TestSuiteOrdering() {
         suite.addTestSuite(MavenIT0010DependencyClosureResolutionTest.class);
         suite.addTestSuite(MavenIT0009GoalConfigurationTest.class);
         suite.addTestSuite(MavenIT0008SimplePluginTest.class);
+        suite.addTestSuite(MavenITmng7344EffectivePomPluginTest.class);
```
gnodet, line 780: New tests are supposed to be at the top of the list, not the bottom.

core-it-suite/src/test/java/org/apache/maven/it/MavenITmng7344EffectivePomPluginTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.File;
+
+import org.apache.maven.shared.verifier.Verifier;
+import org.apache.maven.shared.verifier.util.ResourceExtractor;
+import org.junit.jupiter.api.Test;
+
+public class MavenITmng7344EffectivePomPluginTest extends AbstractMavenIntegrationTestCase {
+    public MavenITmng7344EffectivePomPluginTest() {
+        super("[4.0.0-alpha-13,)");
```
gnodet, line 29: This should be `[4.0.0-beta-4-SNAPSHOT,)`

PR #379, opened 2024-10-06
[MNG-7982] IT’s for transitive dependency management

DidierLoiseau:
Main PR: https://github.com/apache/maven/pull/1785

This PR provides integration tests for [MNG-7982](https://issues.apache.org/jira/browse/MNG-7982), implemented in apache/maven#1357, as well as the legacy (Maven 2/3) behavior.

Unfortunately **these tests fail** with Maven 4.0.0-beta-4 and the current master:
- `MavenITmng7982DependencyManagementTransitivityTest`: the test shows that, currently, the transitive dependency management fails for module `d`, but if you comment it out, you’ll see that it works for `e`, which is inconsistent – dependency management is ignored for just the first level of dependencies, not deeper layers.
- `MavenITmng4720DependencyManagementExclusionMergeTest` was updated to show the behavior expected after MNG-7982: exclusions from dependency management in transitive dependencies should also be picked up.

Both tests can be fixed by making `MavenSessionBuilderSupplier#getDependencyManager(boolean)` return a `TransitiveDependencyManager` when the parameter is `true`, without breaking any other tests.

See also [my comment in MNG-5761](https://issues.apache.org/jira/browse/MNG-5761?focusedCommentId=17887084&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17887084) about this inconsistency.


core-it-suite/src/test/java/org/apache/maven/it/MavenITmng4720DependencyManagementExclusionMergeTest.java
```diff
@@ -39,11 +40,13 @@ public MavenITmng4720DependencyManagementExclusionMergeTest() {
     /**
```
DidierLoiseau, line 40: Should I update the `super()` above? Maybe keep the old test for `[2.0.6,4.0.0)` and move these changes to a new test for `[4.0.0,)`?

I don’t know how this test suite is used for previous releases…

gnodet: It's not really used anymore. At least, we branched for Maven 3.x, not sure how we'll handle things for 4.0, 4.1, etc..., but I was actually considering merging ITs into maven to ease.  I don't really see any good reason not to do that.

gnodet:
@DidierLoiseau could you align the branch names for [this PR](https://github.com/apache/maven-integration-testing/pull/379) and https://github.com/apache/maven/pull/1785, this will allow running #1785 using the new ITs, else it will use master.

DidierLoiseau:
> @DidierLoiseau could you align the branch names for [this PR](https://github.com/apache/maven-integration-testing/pull/379) and [apache/maven#1785](https://github.com/apache/maven/pull/1785), this will allow running #1785 using the new ITs, else it will use master.

I see you have created #384 and apache/maven#1788, so I guess this isn’t necessary anymore?

gnodet:
> > @DidierLoiseau could you align the branch names for [this PR](https://github.com/apache/maven-integration-testing/pull/379) and [apache/maven#1785](https://github.com/apache/maven/pull/1785), this will allow running #1785 using the new ITs, else it will use master.
> 
> I see you have created #384 and [apache/maven#1788](https://github.com/apache/maven/pull/1788), so I guess this isn’t necessary anymore?

Right

PR #381, opened 2024-10-07
Two new ITs for MNG-8288 and MNG-8293

cstamas:
Both problems happens during reactor load up, so ITs code is simple, just invoke "mvn validate" while the actual trick is in resources (projects).


gnodet:
Branch renamed in #382

PR #410, opened 2025-08-29
feat: enable prevent branch protection rules

sebtiem:


.asf.yaml
```diff
@@ -2,6 +2,10 @@
 github:
   description: "Apache Maven core ITs"
   homepage: https://maven.apache.org/core-its
+  protected_branches:
+    master: { }
```
Bukama, line 6: We got several maven-x.y branches which afaik are relevant. @cstamas @gnodet

PR #435, opened 2026-08-23
Build project by JDK 26

slawekjaranowski:
- Add Java 26 to CI matrix
- Bump gmavenplus-plugin to 4.3.1
- Upgrade Groovy dependencies to 4.0.33
- Update test keystore to newer algorithms and 2048 bits


cstamas (approved):
Is good, but with 26 onboard, do we really want to keep 25 in matrix? Runners are already scarce and many times we need to wait a lot for one available...

slawekjaranowski:
> Is good, but with 26 onboard, do we really want to keep 25 in matrix? Runners are already scarce and many times we need to wait a lot for one available...

As this project has not many changes and PRs it is not such important.
But I agree we need minimalize build matrix, I will try to discuss it.

